PR #18, opened 2019-04-12
Add scatter plot type with symbol scale

### LINKED ISSUE #14: Add scatterplot layer

ebb-tide:
Closes #14

### DIFF
--- a/docs/src/components/examples/Scatter.tsx
+++ b/docs/src/components/examples/Scatter.tsx
@@ -1,0 +1,31 @@
+import * as React from 'react'
+import {FunctionComponent} from 'react'
+import {RouteComponentProps} from '@reach/router'
+
+import {Plot, Config} from '../../../../src/'
+
+import {TABLE, COLOR_SCHEMES} from './'
+
+type Props = RouteComponentProps
+
+export const ScatterExample: FunctionComponent<Props> = () => {
+  const config: Config = {
+    table: TABLE,
+    yDomain: [0, 50],
+    layers: [
+      {
+        type: 'scatter',
+        x: '_time',
+        y: '_value',
+        fill: ['cpu'],
+        symbol: ['cpu'],
+        colors: COLOR_SCHEMES.find(d => d.name === 'InfluxDB A').colors,
+      },
+    ],
+  }
+  return (
+    <div className="example-plot">
+      <Plot config={config} />
+    </div>
+  )
+}
--- a/src/components/ScatterLayer.tsx
+++ b/src/components/ScatterLayer.tsx
@@ -1,0 +1,128 @@
+import * as React from 'react'
+import {useRef, useLayoutEffect, useMemo, FunctionComponent} from 'react'
+import {symbol as d3Symbol, SymbolType} from 'd3-shape'
+
+import {Table, Scale, ScatterLayerConfig} from '../types'
+import {PlotEnv} from '../utils/PlotEnv'
+import {clearCanvas} from '../utils/clearCanvas'
+import {GROUP_COL_KEY} from '../constants'
+
+type ScatterData = Array<{
+  xs: number[] | Float64Array
+  ys: number[] | Float64Array
+  fill: string
+  symbol: SymbolType
+}>
+
+const collectScatterData = (
+  table: Table,
+  xColKey: string,
+  yColKey: string,
+  fillScale: Scale<string, string>,
+  symbolScale: Scale<string, SymbolType>
+): ScatterData => {
+  const xCol = table.columns[xColKey].data
+  const yCol = table.columns[yColKey].data
+  const groupCol = table.columns[GROUP_COL_KEY].data as string[]
+  const groupKeys = new Set(groupCol)
+
+  const result = []
+
+  for (let groupKey of groupKeys) {
+    let xs = []
+    let ys = []
+
+    for (let i = 0; i < table.length; i++) {
+      if (groupCol[i] !== groupKey) {
+        continue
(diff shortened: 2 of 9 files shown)


src/utils/getSymbolScale.ts
```diff
@@ -0,0 +1,31 @@
...
+// TODO pass in the symbols you'd like to use, as in color array.
+export const getSymbolScale = (
+  table: Table,
+  symbolColKeys: string[]
+): Scale<string, SymbolType> => {
+  if (!symbolColKeys.length) {
+    return (_i: string) => symbolCircle
+  }
+
+  const groupKeyCol = table.columns[GROUP_COL_KEY]
```
chnn: Hm it seems like this is using the group column computed in the stat, except that group column is based off of the `fill` mapping instead of the `symbol` mapping.

So I think if the layer was specified like:
```
{
  type: 'scatterplot',
  fill: ['a', 'b'],
  symbol: ['c'],
  // ...
}
```
Then the symbol scale would be based off of the `a` and `b` columns rather than the `c` column, which is what was specified.

We can totally address this in a future PR, which might have been what you were thinking. I think the solution would be to append another group column to the table—so there would be one group column based off of the `fill` columns, and an independent column based off of the `symbol` columns.

src/stats/scatter.ts
```diff
@@ -0,0 +1,26 @@
...
+} from '../types'
+import {getFillScale} from '../utils/getFillScale'
+import {getSymbolScale} from '../utils/getSymbolScale'
+import {appendGroupCol} from '../utils/appendGroupCol'
+
+export const scatterStat = (
+  config: SizedConfig,
+  layer: ScatterLayerConfig
+): {table: Table; mappings: ScatterMappings; scales: ScatterScales} => {
+  const table = appendGroupCol(config.table, layer.fill)
```
chnn: See my comment below about the `fill` / `scale` mappings. I think we might want to be appending a second group column here based off of `layer.fill`.

docs/src/components/examples/Scatter.tsx
```diff
@@ -0,0 +1,31 @@
...
+import {TABLE, COLOR_SCHEMES} from './'
+
+type Props = RouteComponentProps
+
+export const ScatterExample: FunctionComponent<Props> = () => {
+  const config: Config = {
+    table: TABLE,
+    yDomain: [0, 50],
+    layers: [
+      {
```
chnn: Hell yeah this looks really great.

src/components/ScatterLayer.tsx
```diff
@@ -0,0 +1,128 @@
...
+  symbolScale: Scale<string, SymbolType>
+): ScatterData => {
+  const xCol = table.columns[xColKey].data
+  const yCol = table.columns[yColKey].data
+  const groupCol = table.columns[GROUP_COL_KEY].data as string[]
+  const groupKeys = new Set(groupCol)
+
+  const result = []
+
+  for (let groupKey of groupKeys) {
```
chnn: There's a minor optimization we can make here that I realized the other day. Instead of mapping over the table once for every group key, we can map over the table once and pick out points for a group key as we see them.

Here's a very similar change made to the line graph: https://github.com/influxdata/vis/commit/acc6ed3b711413e074c37af1bf59895b5d8c4bd4

I've been thinking of this and a lot of the other optimizations in big O terms, which seem particularly relevant for this library, since our performance problem is usually "the data set is too big". If _m_ is the cardinality of the plot and  _n_ is the length of the passed table, the optimization makes this _O(n)_ rather than _O(mn)_. So it's good for high cardinality plots.

src/components/ScatterLayer.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const symbolScale = env.getScale(layerIndex, 'symbol')
+
+  const layerConfig = env.config.layers[layerIndex] as ScatterLayerConfig
+
+  const {x: xColKey, y: yColKey} = layerConfig
+  const {xScale, yScale, innerWidth: width, innerHeight: height} = env
+
+  const scatterData = useMemo(
+    () => collectScatterData(table, xColKey, yColKey, fillScale, symbolScale),
+    [table, xColKey, yColKey]
```
chnn:
```suggestion
    [table, xColKey, yColKey, fillScale, symbolScale]
```

src/components/ScatterLayer.tsx
```diff
@@ -0,0 +1,128 @@
...
+  useLayoutEffect(() => {
+    drawPoints({
+      canvas: canvas.current,
+      scatterData,
+      width,
+      height,
+      xScale,
+      yScale,
+    })
+  })
```
chnn:
```suggestion
  }, [canvas.current, scatterData, width, height, xScale, yScale])
```

chnn: Maybe? I can't tell if memoization will help or hurt here. My gut feeling is that it will help, but I don't know for sure.

src/components/ScatterLayer.tsx
ebb-tide: thanks!

src/utils/getSymbolScale.ts
ebb-tide: I am a little fuzzy on how we manipulate/interpret the columns, if you think it's ok I'll merge this now, and make an issue for myself to understand this properly.

src/stats/scatter.ts
ebb-tide: will return to this as discussed ^^

PR #42, opened 2019-05-09
Fix scatterplot symbol scale bug

ebb-tide:


src/stats/scatter.ts
```diff
@@ -5,22 +5,27 @@ import {
...
-import {appendGroupCol} from '../utils/appendGroupCol'
+import {appendGroupingCol} from '../utils/appendGroupingCol'
+import {SYMBOL_COL_KEY, FILL_COL_KEY} from '../constants'
 
 export const scatterStat = (
   config: SizedConfig,
   layer: ScatterLayerConfig
 ): {table: Table; mappings: ScatterMappings; scales: ScatterScales} => {
-  const table = appendGroupCol(config.table, layer.fill)
+  let table = appendGroupingCol(config.table, layer.fill, FILL_COL_KEY)
```
chnn: Maybe composing functions is clearer? 
```
const table = appendGroupingCol(
  appendGroupingCol(config.table, layer.fill, FILL_COL_KEY),
  layer.symbol,
  SYMBOL_COL_KEY
)
```
eh w/e

src/utils/appendGroupCol.ts
```diff
@@ -16,7 +16,7 @@ export const appendGroupCol = <T extends Table>(
     ...table,
     columns: {
       ...table.columns,
-      [GROUP_COL_KEY]: {data, type: 'string'},
+      [GROUP_COL_KEY]: {data, type: 'string', name: GROUP_COL_KEY},
```
chnn: woops, thanks for fixing this

stories/data.ts
```diff
@@ -3020,6 +3020,1012 @@ export const TABLE: Table = {
       type: 'string',
       name: 'cpu',
     },
+    host: {
```
chnn: yay! I would love to add more data sets too eventually, besides for just the CPU data set. [vega-datasets](https://github.com/vega/vega-datasets) might be useful.

stories/helpers.tsx
```diff
@@ -4,6 +4,7 @@ import {select, text, boolean} from '@storybook/addon-knobs'
 import {Table, isNumeric} from '../src'
 import {TABLE} from './data'
 import * as colorSchemes from '../src/constants/colorSchemes'
+import {includes} from 'lodash'
```
chnn: Weird... `lodash` isn't specified in the `package.json` file. It must be a transitive dependency and somehow importing it works.

Either way, can we use [`Array.prototype.includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) instead?

src/utils/getFillScale.ts
```diff
@@ -25,6 +26,17 @@ export const getFillScale = (
   return getColorScale(Object.keys(domain), colors)
 }
 
+export const getFillScaleForScatter = (
```
chnn: This is great, I want to use this style for the line layer too. Want to make that change in this PR? If not I can do it separately.

ebb-tide: I just wanted to you to look at it first, but I can totally make that change in this PR, no problem.

stories/helpers.tsx
ebb-tide: sure!

PR #44, opened 2019-05-10
Refactor derivation of layer data

chnn:
## The problem

The vis library exposes a `Plot` component which takes a potentially large and nested config object. From this config object we derive the data and code necessary to actually draw a plot.

For example, in order to draw a plot, we need to know the x domain of the data that we are plotting. So we look at each layer in the config object, and what column it maps to the x aesthetic, and compute the x domain by finding the extent across all of those columns in the passed `table`.

We derive a lot of data similarly from the config object. Scales, transformed tables, ticks, margins, and more are all computed by picking fields off of the config object and passing them to functions.

This is nice and simple.

However, the `Plot` often needs to rerender itself with an updated config object. When a config object is updated, we don't want to rederive all of the necessary prerequisites for a plot. For example, if only the color scheme used for a layer changes, we don't need to recompute the x domain.

In fact, since some data is really slow to derive from the config, we don't want to derive it unless absolutely necessary.

Currently, the general pattern we use to avoid unnecessary work is to derive all config data through a `PlotEnv` class. The `PlotEnv` class exposes getters for derived data like the `xDomain`, which internally use memoized/caching/stateful versions of potentially slow utility helper functions like `getXDomainExtent`.

This works pretty well, but there are two issues with the existing implementation. 

1. The first is that not all derived data is cached through memoized helper functions. We attempt to eagerly derive some data such as layer-specific table transforms every time the config is set ([example](https://github.com/influxdata/vis/blob/b8638c90878c07f6581bd93292b8b7bdd49b8ce7/src/utils/PlotEnv.ts#L82-L89)). Determining when the inputs to the derivation are stale or not is difficult and error prone. We don't derive the data lazily through memoized helper functions due to a second issue.

2. The second issue is that the memoization logic for a function ([memoize-one](https://github.com/alexreardon/memoize-one)) is based on the reference identity of inputs to that function, rather than their logical identity. Since it's quite natural as a user of the `Plot` API to update the reference identity of fields in a config without changing their logical identity, this means that the memoization logic is often ineffective—functions recompute even when there are no meaningful changes.

For example, a user of the `Plot` might have a component that looks like this:

```
const MyTerrificPlot = ({table, unrelatedProp}) => {
  const config = {
    table,
    layers: [{
      type: 'line',
      x: 'time',
      y: 'temperature'
      colors: ['blue']
    }]
  }

  return <Plot config={config} />
}
```

Every time this component renders (say, because the `unrelatedProp` changed), the `colors` array is recreated and its reference identity changed. This causes the color scale for the layer to be recomputed. However, this is unnecessary, since the `colors` array is always `['blue']`; in other words, its logical identity hasn't changed.

We could expect that users of the `Plot` component memoize their config fields appropriately. But this is a huge burden, and difficult to do correctly. The vis library should be smarter internally about when to rederive data.

There exists a lot of prior art for solving this problem. The Vega folks need to solve a problem identical to ours: accepting changes to a config object and efficiently rerendering a visualization. They solve the problem by compiling the config into a data flow graph (see [[1]](https://idl.cs.washington.edu/files/2015-ReactiveVega-InfoVis.pdf) and [[2]](https://observablehq.com/@vega/how-vega-works)).

This is an powerful and elegant solution. We should consider adopting it at some point. But it's a lot to engineer—this PR has a more humble solution that was achieved in just a few days.

Similar problems are solved in reactive rendering engines like React and Glimmer. The lastest Glimmer VM has an [impressive architecture](https://github.com/glimmerjs/glimmer-vm/blob/master/guides/01-introduction.md) for a computed property system, which essentially achieves what we are looking for here. Another project that might provide inspiration, inspired in part by Glimmer, is the [Adapton](http://adapton.org/) library.

I hope to study these all more thoroughly in the future.

## The solution

This PR takes a low-tech approach to solving these memoization issues, only tweaking the current `PlotEnv` approach slightly.

To solve the second problem mentioned above, we introduce a ridiculous utility called `identityMerge`. It's job is to take a previous and new config object, and merge them together while preserving the reference identity of matching fields whenever possible. This means that we can happily use these fields as inputs to memoized functions, and thus the original design of the `PlotEnv` continues to work as intended.

This also lets us solve the first problem mentioned—we can now derive _all_ data, included layer table transforms, using the memoized helper pattern.


bthesorceror:
@chnn This all looks good, I would like to make sure we measure performance for deriving data from the config object so we have an accurate way to determine complexity vs. performance gains. Also I don't know if it will be performant in this case but another way we might be able to do comparisons is by hashing the objects and comparing those hashes, that is a tactic I have used in the past for similar issues.

PR #50, opened 2019-05-15
Alter value formatting public API

### LINKED ISSUE #37: Support formatting histogram tick values as times
When mapping a column of type `time` to the `x` aesthetic of a histogram layer, the ticks on the x axis should be formatted as time.

Screen Shot 2019-05-01 at 3 27 16 PM

### LINKED ISSUE #40: Some temporal columns in tooltip are not formatted as time
1. Create a plot with no/default `xTickFormatter` and `yTickFormatter` options
2. Use a temporal column as the `fill` grouping for a `line` layer

Values for the temporal column are formatted as numbers in the line tooltip. They should instead be formatter as times.

e.g. `_start` and `_stop` in the following tooltip:

Screen Shot 2019-05-02 at 2 21 19 PM

We might need to rethink the `xTickFormatter` and `xTickFormatter` options to handle this properly—what about other values that require custom formatting in tooltips? For example, large numbers or numbers with units. Perhaps the `xTickFormatter` and `yTickFormatter` should instead be `numberFormatter` and `timeFormatter` options.

chnn:
Previously a consumer of the vis library could supply optional `xTickFormatter` and `yTickFormatter` fields in their config:

```
{
  ...someConfigOptions,
  xTickFormatter: t => new Date(t).getFullYear(),
  yTickFormatter: t => `${Math.round(t)}%`
}
```

This commit replaces these options with a `valueFormatters` option, which is an object. Each key in the object is a column key, and each value in the object is a function that specifies how to format values within that column.

For example:

```
{
  ...someConfigOptions,
  valueFormatters: {
    myValueCol: x => `${Math.round(t)}%`,
    myTimeCol: t => new Date(t).getFullYear()
  }
}
```

Closes #37
Closes #40


src/utils/defaultNumberFormatter.ts
```diff
@@ -0,0 +1,15 @@
+import {format} from 'd3-format'
+
+const MAX_DECIMALS = 4
```
alexpaxton: Should this be configurable?

chnn: Hmm maybe eventually as like a "precision to use when formatting numbers with the default/fallback value formatter" option. Seems a bit specific so I was hesitant to expose it as part of the public API.

I think right now the way for someone to specify how much precision they want would be to specify a value formatter for the column they care about.

PR #59, opened 2019-05-21
Draw shapes instead of using d3-shape in scatter

ebb-tide:


src/utils/getSymbolScale.ts
```diff
@@ -1,16 +1,24 @@
 import {scaleOrdinal} from 'd3-scale'
-import {SymbolType, symbols} from 'd3-shape'
 import {Table, Scale} from '../types'
 import {SYMBOL_COL_KEY} from '../constants'
 
+export enum SymbolType {
```
chnn: Can you make this a union of string literals? I think string literals are friendlier to consumers of the project that don't use TypeScript.
```
export type Symbol = 'circle' | 'triangle' | 'square'
```
It should still provide the same type safety / benefits to us.

stories/config.ts
```diff
@@ -0,0 +1,5448 @@
+export const conf = {
```
chnn: delete me!

src/components/ScatterLayer.tsx
```diff
@@ -1,11 +1,12 @@
 import * as React from 'react'
 import {useRef, useLayoutEffect, useMemo, FunctionComponent} from 'react'
-import {symbol as d3Symbol, SymbolType} from 'd3-shape'
+import {SymbolType} from '../utils/getSymbolScale'
 
 import {Table, Scale, ScatterLayerConfig} from '../types'
 import {PlotEnv} from '../utils/PlotEnv'
 import {clearCanvas} from '../utils/clearCanvas'
 import {FILL_COL_KEY, SYMBOL_COL_KEY} from '../constants'
+import {drawCircle, drawSquare, drawTriangle} from '../utils/drawShapes'
```
chnn: awesome

PR #61, opened 2019-05-24
Add new shapes to scatter repertoire

ebb-tide:
Add tritip, hollow square, ex, and plus to scatter plot shapes
correct the placement of the circle.

### DIFF
--- a/src/utils/drawShapes.ts
+++ b/src/utils/drawShapes.ts
@@ -1,41 +1,112 @@
 export const drawCircle = (
   ctx: CanvasRenderingContext2D,
-  centerX: number,
-  centerY: number,
-  r?: number
+  x: number,
+  y: number,
+  radius: number = 2
 ) => {
-  const radius = r ? r : 2
-  const x = centerX - radius / 2
-  const y = centerY - radius / 2
-
-  const Zero = 0
-  const TwoPi = Math.PI * 2
-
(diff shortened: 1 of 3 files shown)


src/utils/drawShapes.ts
```diff
@@ -20,22 +17,104 @@ export const drawSquare = (
   centerY: number,
   s?: number
```
chnn: What do you think about making these parameters with defaults?

```suggestion
  size: number = 4
```

Doesn't matter much at all, but might make reading the function signature slightly nicer.

src/utils/getSymbolScale.ts
```diff
@@ -8,7 +8,14 @@ export const getSymbolScale = (table: Table): Scale<string, SymbolType> => {
   const symbolCol = table.getColumn(SYMBOL, 'string')
   const uniqueValues = Array.from(new Set(symbolCol))
 
-  const symbolTypes = ['circle', 'triangle', 'square'] as SymbolType[]
+  const symbolTypes = [
```
chnn: I think
```suggestion
  const symbolTypes: SymbolType[] = [
```
would remove the need for the type assertion.

src/utils/drawShapes.ts
ebb-tide: that is better, thanks

ebb-tide:
I will add those two as issues

PR #81, opened 2019-06-05
Fix unnecessary line layer renders

chnn:
The line layer was being re-rendered on every hover event, due to a bug in the memoization of the x and y scales.

Also adds a story for experimenting with large datasets. The following plot has 5 series with 200,000 points each:

Screenshot 2019-06-05T11:07:43

The size of the dataset can be changed by tweaking the parameters in `stories/sin.ts`.

### DIFF
--- a/src/utils/lineData.ts
+++ b/src/utils/lineData.ts
@@ -51,7 +51,7 @@ export const simplifyLineData = (
     const [simplifedXs, simplifiedYs] = simplify(
       xs.map(x => xScale(x)),
       ys.map(y => yScale(y)),
-      1
+      0.5
     )
 
     result[k] = {xs: simplifedXs, ys: simplifiedYs, fill}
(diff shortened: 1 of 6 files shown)


src/utils/useChangedDepsLogger.ts
```diff
@@ -0,0 +1,18 @@
+import {useRef, MutableRefObject} from 'react'
+
+export const useChangedDepsLogger = deps => {
```
ebb-tide: would you mind to providing an example of how to use this in a comment? Looks useful.

src/utils/lineData.ts
```diff
@@ -51,7 +51,7 @@ export const simplifyLineData = (
     const [simplifedXs, simplifiedYs] = simplify(
       xs.map(x => xScale(x)),
       ys.map(y => yScale(y)),
-      1
+      0.5
```
ebb-tide: I'm curious what the effect of changing the epsilon here was.

chnn: [Check it out](https://observablehq.com/@chnn/running-ramer-douglas-peucker-on-typed-arrays) :)

PR #83, opened 2019-06-06
Add tooltip to scatterplot layers

chnn:
hovering


src/components/ScatterHoverLayer.tsx
```diff
@@ -0,0 +1,106 @@
...
+    xScale,
+    yScale,
+    width,
+    height
+  )
+
+  const canvasRef = useRef<HTMLCanvasElement>(null)
+
+  useLayoutEffect(() => {
+    if (!canvasRef.current) {
```
ebb-tide: Could there be a benefit to returning if rowIndices is undefined as well? I can't quite reason about it, the functionComponent will return null, but will it still do the calculations in drawPoints?

src/components/ScatterHoverLayer.tsx
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  const tooltipData = getPointsTooltipData(
+    rowIndices,
+    table,
+    xColKey,
+    yColKey,
+    FILL,
+    env.getFormatterForColumn,
+    [...new Set([...fillColKeys, ...symbolColKeys])],
```
ebb-tide: I like this.. :)

src/components/ScatterHoverLayer.tsx
chnn: Agh definitely confusing. I think it's OK, since the component will return `null` so `canvasRef.current` will be falsey.

PR #86, opened 2019-06-12
Support shading area below lines

chnn:
Kapture 2019-06-12 at 11 42 14

### DIFF
--- a/src/components/LineLayer.tsx
+++ b/src/components/LineLayer.tsx
@@ -35,6 +35,8 @@ export const LineLayer: FunctionComponent<Props> = ({
     lineWidth,
     hoverDimension,
     maxTooltipRows,
+    shadeBelow,
+    shadeBelowOpacity,
   } = layer
 
   const xColData = table.getColumn(xColKey, 'number')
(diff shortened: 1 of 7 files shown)


src/utils/drawLines.ts
```diff
@@ -10,24 +10,50 @@ interface DrawLinesOptions {
...
+  shadeBelow,
+  shadeBelowOpacity,
+  y0,
 }: DrawLinesOptions): void => {
   const context = canvas.getContext('2d')
 
+  if (shadeBelow) {
+    for (const {xs, ys, fill} of Object.values(lineData)) {
+      const areaGenerator = area<number>()
+        .y0(y0) // TODO
```
ebb-tide: what is this TODO?

src/constants/index.ts
```diff
@@ -60,6 +60,8 @@ export const LAYER_DEFAULTS: {[layerType: string]: Partial<LayerConfig>} = {
     colors: DEFAULT_COLOR_SCHEME,
     interpolation: 'linear',
     maxTooltipRows: 24,
+    shadeBelow: false,
```
ebb-tide: I think the masses want this to be true

src/components/LineLayer.tsx
```diff
@@ -69,8 +71,20 @@ export const LineLayer: FunctionComponent<Props> = ({
       lineData: simplifiedLineData,
       interpolation,
       lineWidth,
+      shadeBelow,
+      shadeBelowOpacity,
+      y0: yScale(0),
```
ebb-tide: Should this be `y0:yScale(yDomain[0])` instead? Or do we always want to shade until zero? Also `y0` is a bit hard to understand what it's for, could it be `shadeAboveY`maybe?

src/utils/drawLines.ts
chnn: whoops, it is outdated

src/constants/index.ts
chnn: 😈

src/components/LineLayer.tsx
chnn: Ah yep totally, nice catch. I will rename as well.

PR #88, opened 2019-06-17
Refactor transforms

### LINKED ISSUE #87: Refactor layer transforms
In preparation for:

https://github.com/influxdata/influxdb/issues/13959
https://github.com/influxdata/giraffe/issues/56
https://github.com/influxdata/giraffe/issues/43

The idea is to decouple the high level user-supplied configuration for a layer from a lower-level "spec" that is used when rendering the layer. This enables multiple types of layer configs to target the same type of spec: for example, the histogram and heatmap layers can both be transformed into a rectangle spec and rendered using the same component.

If we add a layer specific to precomputed/Flux histograms, it can also target this same spec. 

In progress on the `refactor-transforms` branch.

chnn:
Closes #87


src/utils/drawRects.ts
```diff
@@ -0,0 +1,79 @@
...
+  const fillCol =
+    table.getColumn(FILL, 'number') || table.getColumn(COUNT, 'number')
+
+  for (let i = 0; i < yMaxCol.length; i++) {
+    if (yMinCol[i] === yMaxCol[i] || xMinCol[i] === xMaxCol[i]) {
+      // Skip 0-length rects
+      continue
+    }
+
+    const x = xScale(xMinCol[i]) + strokePadding / 2
```
121watts: Why is this divided by `2`?

Issue #79, opened 2019-06-04
Can't control tick generation


chnn:
It's desirable to format ticks using a [binary unit prefix](https://en.wikipedia.org/wiki/Binary_prefix) when plotting data whose y-value is measured in bytes. 

But the y-axis generates tick marks at "nice" intervals, i.e. at powers of 10. These intervals are no longer nice when formatting numbers with a binary unit prefix.

For example, here's the same plot with ticks formatted using a decimal unit prefix:

<img width="577" alt="Screen Shot 2019-06-04 at 2 58 24 PM" src="https://user-images.githubusercontent.com/638955/58916731-5d214f00-86d9-11e9-8052-81992af32d58.png">

It has a nice tick of "6G". When formatting the ticks with a binary unit prefix, it looks like this:

<img width="575" alt="Screen Shot 2019-06-04 at 2 58 31 PM" src="https://user-images.githubusercontent.com/638955/58916780-7aeeb400-86d9-11e9-949d-2397204c4e30.png">

The generated tick is now "5.5879G" (6 gigabytes =~ 5.88 gibibytes), which is no longer nice.

TCL735:
**Requirements:**
- value axis ticks should allow a step and cover a range using that step

- labeling on the value axis should use the same step as value axis

- tooltip's **_value** unit can be a different scale than the tick labeling. _Example_: megabytes (or mebibytes) in the tooltip, while gigabytes (or gibibytes) on the tick labels

TCL735:
The previous comment's 2nd point on having different scales for the value axis and tooltip should be broken down into two separate stories.

In this story, we will focus on changes to the axes. A separate story can focus on changes to the tooltip.

TCL735:
Analysis:

Currently, Giraffe allows `xTicks` and `yTicks` to be specified. When either property is included in the config, that axis is not calculated by Giraffe. When the specified `xTicks` and `yTicks` are not in domain on the rendered axis, then that axis becomes unlabeled.

We will keep the above behavior and add two additional properties per axis that is mutually exclusively with the above.

**New:**
`xTickStep` and `xTotalTicks` are mutually exclusive with and subservient to `xTicks`
`yTickStep` and `yTotalTicks` are mutually exclusive with and subservient to `yTicks`

TCL735:
There is an existing bug that can be fixed with this issue.

[...]

Issue #67, opened 2019-05-28
add ability to add prefix/suffixes to date type axes


ebb-tide:
Currently adding prefix/suffix forces axes labels to be interpreted as strings. Need to figure out a way for date/time formatting to play nicely with time labels

TCL735:
This is possible with the `valueFormatters` config option and `_time` property.

PR #94, opened 2019-07-02
Improve value formatting API, add time zone support

### LINKED ISSUE #80: Add library of tick formatters
There are two issues with the current tick formatting API: #67, #79.

To fix these problems I propose we expose several tick formatting utilities in the vis library, which can handle common cases like formatting times or formatting quantities of bytes.

These utilities could be composed with user specific functions to fix #67.

It would also let us generate ticks more intelligently when the baked-in tick formatting utilities are used, since we can take into account the formatting when generating ticks. This would fix #79.

chnn:
timezones

Suppose a user has a basic Giraffe config:

```
import {Config} from '@influxdata/giraffe'

// ...

const config: Config = {
  table,
  layers: [{
    type: 'line',
    x: 'time',
    y: 'cpuPercentage'
  }]
}
```

Previously, they could control how values from their `table` were displayed using the `valueFormatters` option:

```
import {Config} from '@influxdata/giraffe'

// ...

const config: Config = {
  table,
  valueFormatters: {
    cpuPercentage: x => `${x}%`
  },
  layers: [{
    type: 'line',
    x: 'time',
    y: 'cpuPercentage'
  }]
}
```

However, there were some issues with this:

- It was difficult to make small changes to Giraffe's default formatting of date time values (#67)
- We would generate ugly axis tick marks when a user used custom value formatting (#79)

This PR solves the first issue and enables a solution for the second, while also introducing support for modifying the time zone of a plot. It does this by introducing public utilities for creating formatter functions:

```
import {timeFormatter, Config} from '@influxdata/giraffe'

// ...

const config: Config = {
  table,
  valueFormatters: {
    time: timeFormatter({timeZone: 'America/Chicago'}),
    cpuPercentage: x => `${x}%`
  },
  layers: [{
    type: 'line',
    x: 'time',
    y: 'cpuPercentage'
  }]
}
```

The formatting functions created by utilities like `timeFormatter` are plain functions, but they have also been tagged with a `_GIRAFFE_FORMATTER_TYPE` property. Internally we can recognize such functions and treat them specially when generating ticks. A user can still pass a custom function for format arbitrary columns in their table.

Besides for `timeFormatter`, a `binaryPrefixFormatter` utility and `siPrefixUtility` utility are also introduced. In the future, I think we could add more such utilities (for example, for time durations or imperial distances).

I also added some snapshot tests to catch any regressions in time zone support or binary prefix formatting.

Closes #80


src/utils/formatters.ts
```diff
@@ -0,0 +1,199 @@
...
+
+export const TIME_FORMATTER_TYPE: 'TIME' = 'TIME'
+
+interface TimeFormatter extends Formatter {
+  (x: number, options?: {domainWidth?: number}): string
+
+  _GIRAFFE_FORMATTER_TYPE: typeof TIME_FORMATTER_TYPE
+}
+
+interface TimeFormatterFactoryOptions {
```
chnn: We are using the browser [`Intl`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat) API for zero-dependency time zone and locale support.

src/utils/getTicks.ts
```diff
@@ -1,37 +1,26 @@
...
-    orientation === 'vertical' ? charMetrics.height : charMetrics.width
+  switch (formatter._GIRAFFE_FORMATTER_TYPE) {
+    case 'TIME':
+      return getTimeTicks(domain, rangeLength, numTicks)
 
-  const numTicks = getNumTicks(sampleTick, length, charLength)
-
-  if (columnType === 'time') {
-    return getTimeTicks(domain, length, numTicks)
+    default:
```
chnn: We can solve https://github.com/influxdata/giraffe/issues/79 here by adding another case statement.

src/utils/formatters.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    // 24 hour time when using UTC
+    expect(utcFormatter(d)).toEqual('1/1/2019, 00:00 UTC')
+
+    // Default (locale specific) when not UTC
+    expect(nonUTCFormatter(d)).toEqual('12/31/2018, 4:00 PM PST')
+  })
+})
+
+describe('siPrefixFormatter', () => {
```
ebb-tide: Above and beyond! :)

src/components/Axes.tsx
```diff
@@ -58,6 +62,7 @@ export const drawAxes = ({
 
   const context = canvas.getContext('2d')
   const xAxisY = height - margins.bottom
+  const xDomainWidth = xDomain[1] - xDomain[0]
```
ebb-tide: nice! just checking there's no chance that xDomain[1] < xDomain[0]?

chnn: hmm good point, I think that might be able to happen. I'll update this

PR #99, opened 2019-07-15
Add custom layer support

chnn:
This PR adds a few small features to giraffe to facilitate https://github.com/influxdata/influxdb/pull/14347.

I took the opportunity to clean up the brush event handling code, which was confusing.

I also added the `displayName` prop to all components. This makes them show up with nice names in the React devtools, even when the actual function component names have been mangled by terser (e.g. show `SizedPlot` instead of `n`).

It might be easiest to review commit by commit.

### DIFF
--- a/stories/customLayer.stories.tsx
+++ b/stories/customLayer.stories.tsx
@@ -1,0 +1,43 @@
+import * as React from 'react'
+import {storiesOf} from '@storybook/react'
+
+import {Config, Plot} from '../src'
+
+import {PlotContainer} from './helpers'
+import {CPU} from './data'
+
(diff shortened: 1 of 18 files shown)


stories/customLayer.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const config: Config = {
+    table: CPU,
+    layers: [
+      {
+        type: 'scatter',
+        x: '_time',
+        y: '_value',
+      },
+      {
+        type: 'custom',
```
ebb-tide: looks great!

PR #108, opened 2019-09-20
Mouse position should be aware of y position just like x position

ebb-tide:
Layer legends were visible event when mouse y position was past the height of the canvas. 
With this, hover actions will not be active when mouseY > height.


alexpaxton:
1000%

PR #110, opened 2019-11-11
fix: date weirdness for overlapping dates

Asalem1:
Closes #15808

https://github.com/influxdata/influxdb/issues/15808

### Problem

Date range ticks weren't being updated when values would overlap

### Solution

Updated the way the number of ticks were being calculated and modified the output data due to inconsistent behavior with D3

### DIFF
--- a/giraffe/src/utils/getTicks.ts
+++ b/giraffe/src/utils/getTicks.ts
@@ -3,8 +3,6 @@ import {ticks} from 'd3-array'
 
 import {Formatter, FormatterType} from '../types'
 
-const TICK_DENSITY = 0.5
-
 export const getTicks = (
   domain: number[],
   rangeLength: number,
@@ -13,7 +11,6 @@ export const getTicks = (
 ): number[] => {
   const sampleTick = formatter(domain[1])
   const numTicks = getNumTicks(sampleTick, rangeLength, charLength)
-
   switch (formatter._GIRAFFE_FORMATTER_TYPE) {
     case FormatterType.Time:
       return getTimeTicks(domain, rangeLength, numTicks)
@@ -29,9 +26,7 @@ const getNumTicks = (
   charLength: number
 ): number => {
   const sampleTickWidth = sampleTick.length * charLength
-  const numTicks = Math.round((length / sampleTickWidth) * TICK_DENSITY)
-
-  return Math.max(numTicks, 2)
+  return Math.round(length / sampleTickWidth)
 }
 
 const getTimeTicks = (
@@ -39,9 +34,15 @@ const getTimeTicks = (
   length: number,
   numTicks: number
 ): number[] => {
-  return scaleUtc()
+  const results = scaleUtc()
     .domain([d0, d1])
     .range([0, length])
     .ticks(numTicks)
     .map(d => d.getTime())
+  if (results.length > numTicks) {
+    // added this to force D3 to use the numTicks since
+    // D3 treats the numTicks as a suggestion rather than a source
+    return results.slice(0, numTicks)
+  }
+  return results
 }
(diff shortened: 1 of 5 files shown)


giraffe/src/utils/getTicks.ts
```diff
@@ -29,19 +26,23 @@ const getNumTicks = (
...
   length: number,
   numTicks: number
 ): number[] => {
-  return scaleUtc()
+  const results = scaleUtc()
     .domain([d0, d1])
     .range([0, length])
     .ticks(numTicks)
     .map(d => d.getTime())
+  if (results.length > numTicks) {
```
hoorayimhelping: non blocker: something that would be super helpful would be a comment explaining what it means or represents when `results.length > numTicks` - i can't quite figure it out from local context without making an educated guess or looking up what `scaleUtc` does.

giraffe/src/utils/getTicks.ts
```diff
@@ -29,19 +26,23 @@ const getNumTicks = (
   charLength: number
 ): number => {
   const sampleTickWidth = sampleTick.length * charLength
-  const numTicks = Math.round((length / sampleTickWidth) * TICK_DENSITY)
-
-  return Math.max(numTicks, 2)
+  return Math.round(length / sampleTickWidth)
```
hoorayimhelping: it seems really easy for `sampleTickWidth` to become 0 - someone could pass in 0 to `charLength`, and then this would make some undefined behavior. if that's the case, we should definitely try to guard against that happening:

```typescript
// or find a better default than 0
return sampleTickWidth !== 0 ? Math.round(length / sampleTickWidth) : 0
```

giraffe/src/utils/getTicks.ts
Asalem1: @hoorayimhelping do you mean to move this comment outside of the if?

https://github.com/influxdata/giraffe/pull/110/files#diff-325e130ef8529ac0832abade9a2f50bdR43-R44

giraffe/src/utils/getTicks.ts
Asalem1: I'm not sure how the sampleTickWidth would become 0.  The `sampleTick` that is being based in is a timestamp:

https://github.com/influxdata/giraffe/pull/110/files#diff-325e130ef8529ac0832abade9a2f50bdR12

hoorayimhelping: @Asalem1
>I'm not sure how the `sampleTickWidth` would become 0. The `sampleTick` that is being based in is a timestamp:

`sampleTick` is all good, you're absolutely right! the issue is  `sampleTick` is [multiplied by `charLength`](https://github.com/influxdata/giraffe/blob/c20c75449edc745adc5555ff6fde2c195658e280/giraffe/src/utils/getTicks.ts#L28):
```ts
const sampleTickWidth = sampleTick.length * charLength
```
which is [passed into `getNumTicks`](https://github.com/influxdata/giraffe/blob/c20c75449edc745adc5555ff6fde2c195658e280/giraffe/src/utils/getTicks.ts#L13) 

```ts
const numTicks = getNumTicks(sampleTick, rangeLength, charLength)
```

by [outside input:](https://github.com/influxdata/giraffe/blob/c20c75449edc745adc5555ff6fde2c195658e280/giraffe/src/utils/getTicks.ts#L9)

```ts
export const getTicks = (
  domain: number[],
  rangeLength: number,
  charLength: number,
  formatter?: Formatter
): 
```
and so it's pretty untrustworthy at this level.

giraffe/src/utils/getTicks.ts
hoorayimhelping: i guess it's this: i see that we're sending the results array up to `numTicks` to appease d3, i just don't understand why `numTicks` would be greater than `results.length` or what it really means for the state of the app

PR #116, opened 2019-11-18
fix(undefined-timezone): fixed issue where 24h time formats were displaying undefined timezones

Asalem1:
Closes #15808:

https://github.com/influxdata/influxdb/issues/15808

### Problem

The default statement for formatting the date/time in giraffe is based on US 12h timezones. When users from countries that have a 24h based system access giraffe, their timezone is undefined.

### Solution

Updated default format to be the 24h system. This is a patch solution until a new feature is implemented that allows users to specify their display time preferences

### DIFF
--- a/giraffe/src/utils/formatters.ts
+++ b/giraffe/src/utils/formatters.ts
@@ -167,13 +167,15 @@ export const timeFormatter = ({
         timeFormat = timeFormats.zoned12
       } else if (hour12 === false) {
         timeFormat = timeFormats.local24
-      } else {
-        // this is a temporary solution. Russ said that we should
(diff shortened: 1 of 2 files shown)


Asalem1:
> isn't the error that the timeZone being passed in from influxdb is incorrect?

So the timezone that's being passed is actually based on a person's local location, which is correct. The error is that the format of the locale (i.e. en-US, en-GB, etc...) is currently **not** being set on influxdb (a feature that will be implemented in the future)

Asalem1:
> Looks like there's a big hole in the intl date implementation where we have to sniff for this data instead of querying for it. 👍

Agreed

Issue #43, opened 2019-05-09
Support stacked position option in line layers


chnn:
Similar to the `position` option in a histogram layer config, we should support a `position` option in line layer configs.

```
type LinePosition = 'overlaid' | 'stacked'
```

It should default to `'overlaid'`.

The stacking logic should take place in a stat. We may have to refactor the `collectLineData` and `simplifyLineData` utils in the line layer to make this all work out.

chnn:
I did some investigation.

- We can either build the stacking transform ourselves, or [use d3](https://github.com/d3/d3-shape#stacks)
- d3 outputs data in a row-oriented format that will be a pain to integrate with our current `PlotEnv` / `Table` interfaces
- There's [a lot of fantastic parameters to a stacked line graph](http://leebyron.com/streamgraph/stackedgraphs_byron_wattenberg.pdf) that encourage different visualization insights. I would love to expose some of these parameters in the line layer config

TCL735:
Building our own stacking transform is the easiest way to go.

Separately, other challenges to consider:
- shading each layer with its own color rather than the resulting of color of the stacked layers
- line hover tooltip will reflect the cumulative value
- do we need a separate column in the hover tooltip for the original single value (non-cumulative)?

PR #119, opened 2019-11-25
feat: add option for stacked line layers

TCL735:
Closes https://github.com/influxdata/giraffe/issues/43

### DIFF
--- a/giraffe/src/transforms/line.ts
+++ b/giraffe/src/transforms/line.ts
@@ -1,13 +1,36 @@
-import {Table, LineLayerSpec, LineData} from '../types'
+import {
+  Table,
+  LineLayerSpec,
+  LineData,
+  LinePosition,
+  DomainLabel,
+} from '../types'
 import {FILL} from '../constants/columnKeys'
 import {createGroupIDColumn, getNominalColorScale} from './'
+
+export const getPreviousDomainNextValue = (
+  lineData: LineData,
+  groupID: number,
+  domainLabel: DomainLabel
+): number => {
+  if (!lineData[groupID] || !lineData[groupID - 1]) {
+    return 0
+  }
+  const currentGroupLength = lineData[groupID][domainLabel].length
+  const previousGroupLength = lineData[groupID - 1][domainLabel].length
+  if (previousGroupLength > currentGroupLength) {
+    return lineData[groupID - 1][domainLabel][currentGroupLength]
+  }
+  return 0
+}
 
 export const lineTransform = (
   inputTable: Table,
   xColumnKey: string,
   yColumnKey: string,
   fillColKeys: string[],
-  colors: string[]
+  colors: string[],
+  position: LinePosition
 ): LineLayerSpec => {
   const [fillColumn, fillColumnMap] = createGroupIDColumn(
     inputTable,
@@ -28,13 +51,17 @@ export const lineTransform = (
   for (let i = 0; i < table.length; i++) {
     const groupID = fillColumn[i]
     const x = xCol[i]
-    const y = yCol[i]
+    let y = yCol[i]
 
     if (!lineData[groupID]) {
       lineData[groupID] = {xs: [], ys: [], fill: fillScale(groupID)}
     }
 
     lineData[groupID].xs.push(x)
+
+    if (position === 'stacked') {
+      y += getPreviousDomainNextValue(lineData, groupID, DomainLabel.Y)
+    }
     lineData[groupID].ys.push(y)
 
     if (x < xMin) {
--- a/giraffe/src/types/index.ts
+++ b/giraffe/src/types/index.ts
@@ -82,12 +82,15 @@ export type LineInterpolation =
 */
 export type LineHoverDimension = 'x' | 'y' | 'xy'
 
+export type LinePosition = 'overlaid' | 'stacked'
+
 export interface LineLayerConfig {
   type: 'line'
   x: string
   y: string
   fill?: string[]
   colors?: string[]
+  position?: LinePosition | 'overlaid'
   interpolation?: LineInterpolation
   hoverDimension?: LineHoverDimension | 'auto'
   lineWidth?: number
@@ -360,4 +363,9 @@ export type LayerSpec = LineLayerSpec | ScatterLayerSpec | RectLayerSpec
 export enum ErrorName {
   UnknownColumnTypeError = 'UnknownColumnTypeError',
   SchemaMismatchError = 'SchemaMismatchError',
+}
+
+export enum DomainLabel {
+  X = 'xs',
+  Y = 'ys',
 }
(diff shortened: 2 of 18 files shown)


TCL735:
Example of stacked line layer. Notice also the tooltip has an additional column named "cumulative" which reflects the value of the hovered points, while the "_value" is the original single values of the points.

Screen Shot 2019-11-25 at 10 37 04 AM

TCL735:
With shading

Screen Shot 2019-11-25 at 10 42 51 AM

giraffe/src/transforms/line.ts
```diff
@@ -1,13 +1,30 @@
...
+  domainLabel: string
+): number => {
+  if (groupID < 1) {
+    return 0
+  }
+  const currentGroupLength = lineData[groupID][domainLabel].length
+  const previousGroupLength = lineData[groupID - 1][domainLabel].length
+  if (previousGroupLength > currentGroupLength) {
+    return lineData[groupID - 1][domainLabel][currentGroupLength]
+  }
```
hoorayimhelping: this block has a couple of potential undefined accesses (`cannot access property 'length' of null`). not sure how certain we can be that this data exists, but if we want to be more guarded we could do something like:
```ts
const currentGroup = lineData[groupID]
const previousGroup = lineData[groupID - 1]

if (!currentGroup || !previousGroup) {
  return 0
}

const currentDomainLabel = currentGroup[domainLabel]
const previousDomainLabel = previousGroup[domainLabel]

if (currentDomainLabel && previousDomainLabel) {
  if (previousDomainLabel.length > currentDomainLabel.length) {
    return previousDomainLabel[currentDomainLabel.length]
  }
}

return 0
```

alternatively, you could be kinda lazy and yolo it (not a fan of this):

```ts
try {
  const currentGroupLength = lineData[groupID][domainLabel].length
  const previousGroupLength = lineData[groupID - 1][domainLabel].length
  if (previousGroupLength > currentGroupLength) {
    return lineData[groupID - 1][domainLabel][currentGroupLength]
  }
} catch () {
  // chomp
  return 0
}
```

giraffe/src/utils/lineData.ts
```diff
@@ -51,3 +51,15 @@ export const simplifyLineData = (
...
+  lineData: LineData,
+  domainLabel: string
+): Array<number> => {
+  let result = []
+  const numberOfLines = Object.keys(lineData).length
+  for (let index = 0; index < numberOfLines; index += 1) {
+    result = result.concat(lineData[index][domainLabel])
+  }
+  return result
+}
```
hoorayimhelping: would love to see even a simple test for this!

giraffe/src/utils/tooltip.ts
```diff
@@ -38,6 +47,25 @@ const getTooltipGroupColumns = (
...
+  const yColData = getDomainDataFromLines(lineData, 'ys')
+  const yFormatter = getValueFormatter(yColKey)
+  return {
+    key: yColKey,
+    name: 'cumulative',
+    type: table.getColumnType(yColKey),
+    colors: rowColors,
+    values: rowIndices.map(i => yFormatter(yColData[i])),
+  }
+}
```
hoorayimhelping: some jest tests for this would be fantastic

giraffe/src/transforms/line.ts
```diff
@@ -1,13 +1,30 @@
-import {Table, LineLayerSpec, LineData} from '../types'
+import {Table, LineLayerSpec, LineData, LinePosition} from '../types'
 import {FILL} from '../constants/columnKeys'
 import {createGroupIDColumn, getNominalColorScale} from './'
 
+const getPrevCumulativeDomainValue = (
+  lineData: LineData,
+  groupID: number,
+  domainLabel: string
```
ebb-tide: I'm surprised this was not typed previously. Would you be willing to try typing 'xs' and 'ys' as an enum?

giraffe/src/types/index.ts
```diff
@@ -82,12 +82,15 @@ export type LineInterpolation =
...
 
+export type LinePosition = 'overlaid' | 'stacked'
+
 export interface LineLayerConfig {
   type: 'line'
   x: string
   y: string
   fill?: string[]
   colors?: string[]
+  position?: LinePosition
```
ebb-tide: do you think it would make sense to provide a default here?

giraffe/src/transforms/line.ts
TCL735: Sure thing. Will add that to the next push.

giraffe/src/types/index.ts
TCL735: Yes, makes perfect sense. I left it out by mistake. Will add that in with the next push.

giraffe/src/transforms/line.ts
TCL735: Yes, I thought a while on this block. `return 0` happening twice did not feel ideal to me and i would like to come up with something more elegant here. I will look into your suggestion as I write the tests.

giraffe/src/utils/lineData.ts
TCL735: Will add the tests with the next push.

giraffe/src/utils/tooltip.ts
TCL735: Will add the tests with the next push.

TCL735:
Aside from the reviews, I noticed upon further testing that sometimes a line's data will not be in the tooltip when the _value is 0. This happens intermittently, as sometimes the tooltip will correctly display the entire line's data. Will investigate this and report back.

TCL735:
The issue with the tooltip missing a line when _value is 0 is due to domain stretching. When the lines are stacked, the Y domain is stretched (as compared to overlaid position) from adding all of the values together. This causes the hover points to be estimated when the number of time ticks are too few. We'll look into adding a possible solution here. If not possible, then a separate issue will be opened.

TCL735:
> Aside from the reviews, I noticed upon further testing that sometimes a line's data will not be in the tooltip when the _value is 0. This happens intermittently, as sometimes the tooltip will correctly display the entire line's data when _value is 0. Will investigate this and report back.

This is a data issue. The sample test data does not have enough records to allow the hover points to be found with high precision when a line has a _value 0 (and thus causing two line layers to intersect when stacked). Increasing the number of records in the sample data makes this issue un-reproduceable. Intersecting lines will have their own hover points.

In real applications, this issue is moot as the data volume is expected to be much higher than the sample test data in storybook.

giraffe/src/transforms/line.ts
```diff
@@ -28,13 +51,17 @@ export const lineTransform = (
   for (let i = 0; i < table.length; i++) {
```
hoorayimhelping: not a blocker, just a thought for the future: might consider converting this to `forEach` function if at all possible and unless there's a strong reason to use a `for` loop - manually iterating is sooooooooo 1989

PR #121, opened 2019-12-02
refactor: remove unnecessary interface extension

TCL735:
Removes unnecessary interface extension.


TCL735:
@russorat not stat buffing

PR #122, opened 2019-12-05
feat(formatTime): added in capacity to output am/pm in custom formats

Asalem1:
### Problem

Currently passing in a custom formatted time into the `formatTime` function will error out if a flag for `a` (am/pm) is passed in the time format 

### Solution

Added a custom check for whether the formatted time includes an `a` and resorts to forcing the format to set the times to 12h with `am/pm`


TCL735:
Looks good.

PR #123, opened 2019-12-06
Feature worldmap

alespour:
Per  Russ's suggestion , this is a WIP PR that contain early prototype of "worldmap" layer, with example storybook included. 

It uses Open Layers (https://openlayers.org) to show maps and features. There are some issue to be solved, such as better React and Open Layers  intergration (use its react wrapper or so), cleaning up etc, and even consideration to use Leaflet library instead (Grafana uses it in its worldmap visualization plugin)


russorat:
@alespour any updates here? as soon as you are ready for a review, let us know.

ivankudibal:
This can be closed w/o merge, there is new component

PR #124, opened 2019-12-06
fix: undefined format check added

Asalem1:


TCL735:
Unit tests are still failing (there is no check for it, must be run locally). We can address with a separate issue.

PR #130, opened 2019-12-07
chore: upgrade eslint and fix errors, upgrade other dependencies

hoorayimhelping:


.eslintrc.json
```diff
@@ -38,6 +38,8 @@
     "no-useless-escape": "off",
     "no-undef": "off",
     "no-fallthrough": "off",
+    "prefer-rest-params": "off",
+    "@typescript-eslint/no-empty-function": "off",
```
TCL735: Why turn this off? Both influxdb and clockface do not have it off. Curious to know if this will cause any issues specifically for giraffe. Perhaps we should remove this to have a consistent style across multiple repos.

hoorayimhelping: @TCL735 I turned it off because it was causing a failure, and I didn't want to modify any code beyond `--fix` when upgrading libraries. The goal is to get the security vulnerabilities patched. We should revisit this style later - I don't have a ton of context on this. I'm just trying to fix critical vulnerabilities.

I wrote up an issue we can address later: https://github.com/influxdata/giraffe/issues/131

hoorayimhelping: @TCL735 I turned it off because it was causing a failure, and I didn't want to modify any code beyond --fix when upgrading libraries. The goal is to get the security vulnerabilities patched. We should revisit this style later - I don't have a ton of context on this. I'm just trying to fix critical vulnerabilities.

I wrote up an issue we can address later: #131

PR #132, opened 2019-12-11
fix: use dependency's 12-hour format when specified

TCL735:
Closes https://github.com/influxdata/influxdb/issues/16122

### DIFF
--- a/giraffe/src/utils/formatters.test.ts
+++ b/giraffe/src/utils/formatters.test.ts
@@ -23,15 +23,29 @@ describe('timeFormatter', () => {
       timeZone: 'UTC',
       format: 'YYYY-MM-DD HH:mm:ss a ZZ',
     })
+    const utcFormatterWithFormatWithLowerH = timeFormatter({
+      timeZone: 'UTC',
+      format: 'YYYY-MM-DD hh:mm:ss a ZZ',
+    })
     const nonUTCFormatterWithFormat = timeFormatter({
       timeZone: 'America/Los_Angeles',
       format: 'YYYY-MM-DD HH:mm:ss a ZZ',
+    })
+    const nonUTCFormatterWithFormatWithLowerH = timeFormatter({
+      timeZone: 'America/Los_Angeles',
+      format: 'YYYY-MM-DD hh:mm:ss a ZZ',
     })
 
     const d = new Date('2019-01-01T00:00Z')
 
     expect(utcFormatterWithFormat(d)).toEqual('2019-01-01 12:00:00 AM UTC')
+    expect(utcFormatterWithFormatWithLowerH(d)).toEqual(
+      '2019-01-01 12:00:00 AM UTC'
+    )
     expect(nonUTCFormatterWithFormat(d)).toEqual('2018-12-31 4:00:00 PM PST')
+    expect(nonUTCFormatterWithFormatWithLowerH(d)).toEqual(
+      '2018-12-31 04:00:00 PM PST'
+    )
   })
 
   test('can format times with format strings', () => {
--- a/giraffe/src/utils/formatters.test.ts
+++ b/giraffe/src/utils/formatters.test.ts
@@ -18,6 +18,22 @@ describe('timeFormatter', () => {
     expect(nonUTCFormatter(d)).toEqual('2018-12-31 04:00:00 PM PST')
   })
 
+  test('uses AM/PM when given "a" in the format regardless of time zone or time format', () => {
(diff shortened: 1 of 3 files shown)


giraffe/src/utils/formatters.ts
```diff
@@ -139,6 +139,7 @@ export const timeFormatter = ({
   // for example, modifying the timezone before determing the `am/pm` will
   // output the timezone incorrectly. The same goes for determining the `HH`, etc...
   const formatStringFormatter = createDateFormatter({
+    hh: ({hour}) => (Number(hour) < 10 ? `0${hour}` : String(hour)),
```
Asalem1: If we're looking to prepend a `0` we can use the `lhour` prop since that should reflect the `24h` hour

giraffe/src/utils/formatters.test.ts
```diff
@@ -18,13 +18,44 @@ describe('timeFormatter', () => {
...
+
   test('can format times with format strings', () => {
     const tests = [
       ['YYYY-MM-DD HH:mm:ss', '2019-01-01 00:00:00'],
       ['YYYY-MM-DD HH:mm:ss ZZ', '2019-01-01 00:00:00 UTC'],
       ['DD/MM/YYYY HH:mm:ss.sss', '01/01/2019 00:00:00.000'],
       ['MM/DD/YYYY HH:mm:ss.sss', '01/01/2019 00:00:00.000'],
       ['YYYY/MM/DD HH:mm:ss', '2019/01/01 00:00:00'],
+      ['hh:mm a', '12:00 AM'],
       ['HH:mm a', '12:00 AM'],
```
Asalem1: We should consider removing this option since their aim is the same

giraffe/src/utils/formatters.ts
```diff
@@ -139,6 +139,7 @@ export const timeFormatter = ({
...
     HH: ({lhour}) => {
       if (format && format.includes('a')) {
```
Asalem1: this might not be necessary now that we've added in the other formatter data

stories/src/index.stories.tsx
```diff
@@ -88,6 +104,22 @@ storiesOf('XY Plot', module)
     const x = xKnob(table)
     const y = yKnob(table)
     const timeZone = timeZoneKnob()
+    const timeFormat = select(
+      'Time Format',
+      {
+        'DD/MM/YYYY HH:mm:ss.sss': 'DD/MM/YYYY HH:mm:ss.sss',
+        'MM/DD/YYYY HH:mm:ss.sss': 'MM/DD/YYYY HH:mm:ss.sss',
+        'YYYY/MM/DD HH:mm:ss': 'YYYY/MM/DD HH:mm:ss',
+        'hh:mm a': 'hh:mm a',
```
Asalem1: have we added this option onto the influxdb formatters or did we replace the `HH:mm a`?

giraffe/src/utils/formatters.ts
TCL735: Prepending 0 is to ensure that hour displays correctly in all environments. Locally, I was seeing "4" for four o'clock, but in CircleCI it was "04", causing tests to pass locally but fail in the pipeline.

Using lhour doesn't solve this problem. Four o'clock PM becomes "16", subtracting 12 gives "4" -- which requires the 0 prepend.

Which brings up a good point: there is now a subtle difference between using `hh:mm a` versus `HH:mm a` -- the leading zero may or may not be there. `hh:mm a` will always display the leading zero. So, now we have an extra option for more subtleties in the formatting.

Asalem1: 👍

giraffe/src/utils/formatters.ts
TCL735: No, let's keep it for now. There are very subtle differences between the two and there might be a use case for this when we build out the options in time formats.

stories/src/index.stories.tsx
TCL735: After giraffe's next release, i will add this to influxdb and bump giraffe at the same time.

giraffe/src/utils/formatters.test.ts
TCL735: Their output is the same. giraffe should be able to handle both formats, and this test checks for that.

Asalem1: for sure, but did we want to have the `HH:mm a` option on the list since it'll output the same thing?

Issue #12, opened 2019-04-10
Enable using a log scale


chnn:
Add a config option to each layer that enables a user to view a plot with a log scale.

PR #134, opened 2019-12-14
feat(giraffe): add log scale type

goller:
Closes: #12 

X and Y Scale can now be updated to be either linear or logarithmic.

Typically, switching to log scale is in response to skewness towards large values.

### DIFF
--- a/giraffe/src/utils/PlotEnv.ts
+++ b/giraffe/src/utils/PlotEnv.ts
@@ -5,7 +5,7 @@ import {extentOfExtents} from './extrema'
 import {identityMerge} from './identityMerge'
 import {MemoizedFunctionCache} from './MemoizedFunctionCache'
 import {timeFormatter} from './formatters'
-import {getLinearScale} from './getLinearScale'
+import {getScale} from './getScale'
 import {lineTransform} from '../transforms/line'
 import {scatterTransform} from '../transforms/scatter'
 import {histogramTransform} from '../transforms/histogram'
@@ -118,10 +118,10 @@ export class PlotEnv {
   }
 
   public get xScale(): Scale<number, number> {
-    const getXScaleMemoized = this.fns.get('xScale', getLinearScale)
+    const scale = getScale(this.config.xScale)
     const {xDomain, rangePadding, innerWidth} = this
 
-    return getXScaleMemoized(
+    return scale(
       xDomain[0],
       xDomain[1],
       rangePadding,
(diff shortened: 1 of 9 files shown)


TCL735:
The xScale and yScale function can each be different functions now. Accordingly, memoization should be applied on all the possible functions. We could add the config option to the memoized key name and that should take care of each of the different possible functions.

giraffe/src/utils/PlotEnv.ts
```diff
@@ -118,10 +118,10 @@ export class PlotEnv {
   }
 
   public get xScale(): Scale<number, number> {
-    const getXScaleMemoized = this.fns.get('xScale', getLinearScale)
+    const scale = getScale(this.config.xScale)
```
TCL735: To memoize, it would be something like this:
<code>const getXScaleMemoized = this.fns.get(&#96;xScale_${this.config.xScale}&#96;, getScale(this.config.xScale))</code>

giraffe/src/utils/PlotEnv.ts
```diff
@@ -118,10 +118,10 @@ export class PlotEnv {
...
     const {xDomain, rangePadding, innerWidth} = this
 
-    return getXScaleMemoized(
+    return scale(
```
TCL735: To keep the `getXScaleMemoized` see above comment

giraffe/src/utils/PlotEnv.ts
```diff
@@ -130,10 +130,10 @@ export class PlotEnv {
   }
 
   public get yScale(): Scale<number, number> {
-    const getYScaleMemoized = this.fns.get('yScale', getLinearScale)
+    const scale = getScale(this.config.yScale)
```
TCL735: To memoize, it would be something like this:
<code>const getYScaleMemoized = this.fns.get(&#96;yScale_${this.config.yScale}&#96;, getScale(this.config.yScale))</code>

giraffe/src/utils/PlotEnv.ts
```diff
@@ -130,10 +130,10 @@ export class PlotEnv {
...
     const {yDomain, rangePadding, innerHeight} = this
 
-    return getYScaleMemoized(
+    return scale(
```
TCL735: To keep `getYScaleMemoized` see above comment

goller:
@ebb-tide @TCL735 I've add memoization; please take a look!

Issue #135, opened 2019-12-17
x-axis should label efficiently and fill the entire axis whenever possible


TCL735:
Add more labels on the x-axis

This
![image](https://user-images.githubusercontent.com/10736577/71024012-d2be9c80-20b8-11ea-8237-46bf0ad17111.png)

should become this
![image](https://user-images.githubusercontent.com/10736577/71024029-df42f500-20b8-11ea-830b-8310568dd396.png)

PR #146, opened 2020-01-07
fix: use as many time ticks as d3 can fit on the x-axis

TCL735:
Closes https://github.com/influxdata/giraffe/issues/135 and https://github.com/influxdata/influxdb/issues/16395

Instead of using 'W' for the average character width, we measure the exact width of the label. Then we ask d3 to give us as many time ticks as will fit on the x-axis with the following restrictions to act as padding around the labels:
- round down when finding the maximum number as the first suggestion to d3
- four or more labels: the sum of all the lengths of the labels is shorter than the available space by at least one label's length
- less than four labels: 0.25, 0.50, 0.75 label length as spacing for one, two, and three labels respectively

### DIFF
--- a/giraffe/src/utils/__mocks__/getTextMetrics.ts
+++ b/giraffe/src/utils/__mocks__/getTextMetrics.ts
@@ -1,3 +1,12 @@
+/*
+  utils/getTextMetrics uses the DOM to measure the actual width and height of a string
+  Since Jest does not have access to the DOM, we will fix the width and height to be some arbitrary numbers,
+  so that our tests will run properly for consumers of getTextMetrics
+
+  This is a mock using a fake font that will always have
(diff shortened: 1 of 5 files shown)


giraffe/src/utils/__mocks__/getTextMetrics.ts
```diff
@@ -0,0 +1,4 @@
+export const getTextMetrics = (...args) => ({
+  width: Math.floor(args[1].length * 1.2),
+  height: 12,
```
ebb-tide: maybe add a comment to explain magic numbers ?

TCL735: Sure thing, will do that.

Issue #145, opened 2020-01-02
Tests: Tooltip


TCL735:
- [x] Add js tests for utils of component `Tooltip`
- [x] Re-order the imports by grouping them by purpose such as `utils`, `constants`, `types`

desa:
@TCL735 is there more to do on this issue?

PR #153, opened 2020-02-06
fix: Use correct scale when hovering over stacked line graphs

TCL735:
Closes https://github.com/influxdata/influxdb/issues/16489 and https://github.com/influxdata/giraffe/issues/145

**Stacked Line Graphs**
- Fixes the data for the hovered points when lines intersect (have similar values)
- Sorts the order of the rows in the tooltip to match the order of the hovered points
- Adds a column in the tooltip to make it clear how many lines are represented in each row
- Adds unit tests for the tooltip utility

### DIFF
--- a/giraffe/src/utils/fixtures/tooltip.ts
+++ b/giraffe/src/utils/fixtures/tooltip.ts
@@ -1,0 +1,49 @@
+import {newTable} from '../newTable'
+
+interface SampleTableOptions {
+  include_negative?: boolean
+  all_negative?: boolean
+  decimalPlaces?: number
+  maxValue?: number
+  numberOfRecords?: number
+  recordsPerLine?: number
+  hoveredRowIndices?: number[]
+}
+
+const turnNegativeRandomly = (num: number) =>
+  num * (Math.floor(Math.random() * 2 + 1) === 1 ? 1 : -1)
+
+const getRandomNumber = (max: number, decimalPlaces: number) =>
+  Number((Math.random() * Math.floor(max)).toFixed(decimalPlaces))
+
+export const columnKey = 'cpu'
+
+export const createSampleTable = (options: SampleTableOptions) => {
+  const {
+    include_negative = false,
+    all_negative = false,
+    decimalPlaces = 2,
+    maxValue = 100,
+    numberOfRecords = 20,
+    recordsPerLine = 5,
+  } = options
+
+  const now = Date.now()
+  const TIME_COL = []
+  const VALUE_COL = []
+  const CPU_COL = []
+
+  for (let i = 0; i < numberOfRecords; i += 1) {
+    let num = getRandomNumber(maxValue, decimalPlaces)
+    if (include_negative) {
+      num = all_negative ? Math.abs(num) * -1 : turnNegativeRandomly(num)
+    }
+    VALUE_COL.push(num)
+    CPU_COL.push(`${columnKey}${Math.floor(i / recordsPerLine)}`)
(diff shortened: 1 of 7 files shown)


stories/src/data/stackedLineLayer.ts
```diff
@@ -10,7 +10,7 @@ const VALUE_COL = []
 const CPU_COL = []
 
 function getRandomInt(max) {
-  return Math.floor(Math.random() * Math.floor(max))
+  return Math.random() * Math.floor(max)
```
ebb-tide: this makes it not an int anymore right?

TCL735: Yes, i will rename the function.

giraffe/src/utils/tooltip.ts
```diff
@@ -7,11 +7,51 @@ import {
...
+const getDataSortOrder = (
+  lineData: LineData,
+  hoveredRowIndices: number[],
+  position: LinePosition
+): number[] => {
+  if (position === 'overlaid') {
+    return hoveredRowIndices
+  }
+  const map = {}
+  const values = Object.keys(lineData).reduce(
```
ebb-tide: why not use map here?

TCL735: `lineData` has multiple properties, each of them is a nested object with some array inside. I am trying to combine all of those arrays into one long array. reduce is good taking many things and turning them into a single thing, so that was my first instinct here.

hoorayimhelping:
Sorry, just realized I had this stupid review sitting here. Good work, just some like super nitpicky stuff - feel free to ignore if disagree

giraffe/src/utils/tooltip.ts
```diff
@@ -7,11 +7,51 @@ import {
...
 const isVoid = (x: any) => x === null || x === undefined
 
+const orderDataByValue = (
+  originalOrder: number[],
+  nextOrder: number[],
+  data: Array<any>
+) => {
+  const map = {}
+  originalOrder.forEach((key, i) => (map[key] = data[i]))
+  return nextOrder.map(index => map[index])
```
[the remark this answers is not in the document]
TCL735: i love nitpicks and i agree here, the variables can have better names. i will make the change.

giraffe/src/utils/tooltip.ts
```diff
@@ -7,11 +7,51 @@ import {
...
+  position: LinePosition
+): number[] => {
+  if (position === 'overlaid') {
+    return hoveredRowIndices
+  }
+  const map = {}
+  const values = Object.keys(lineData).reduce(
+    (accum, value) => accum.concat(lineData[value][DomainLabel.Y]),
+    []
+  )
```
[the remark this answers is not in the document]
TCL735: will change

giraffe/src/utils/fixtures/tooltip.ts
```diff
@@ -0,0 +1,48 @@
...
+  include_negative?: boolean
+  all_negative?: boolean
+  decimalPlaces?: number
+  maxValue?: number
+  numberOfRecords?: number
+  recordsPerLine?: number
+}
+
+const turnNegativeRandomly = (num: number) =>
+  num * (Math.floor(Math.random() * 2 + 1) === 1 ? 1 : -1)
```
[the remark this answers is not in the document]
TCL735: Good point here. I did this quickly and looking at it carefully, it's roughly 50% each. I will improve it to be "random" or at least psuedo-random.

Issue #158, opened 2020-02-14
Stacked line cumulative is incorrect when lines do not cover the same time range


TCL735:
Stacked line cumulative is not counting correctly when separate lines do not cover the same time period.

<img width="1680" alt="image" src="https://user-images.githubusercontent.com/10736577/74556910-e13bff80-4f13-11ea-9ad7-86b0b8166f4a.png">

PR #161, opened 2020-02-21
fix: calculate correct cumulative values for lines of different lengths

TCL735:
Closes https://github.com/influxdata/giraffe/issues/158

By creating a map of all the time points, we can correctly track and add up cumulative values as the time points are traversed during line transformation

### DIFF
--- a/giraffe/src/transforms/line.ts
+++ b/giraffe/src/transforms/line.ts
@@ -21,9 +21,12 @@ export const mapCumulativeValuesToTimeRange = (
       cumulativeValues[time] = {}
     }
     const maxGroup = fillCol[index]
-    for (let i = 0; i < maxGroup; i += 1) {
-      if (!cumulativeValues[time][i]) {
-        cumulativeValues[time][i] = 0
+
+    if (!cumulativeValues[time][maxGroup - 1]) {
+      for (let i = 0; i < maxGroup; i += 1) {
+        if (!cumulativeValues[time][i]) {
+          cumulativeValues[time][i] = 0
+        }
       }
     }
     cumulativeValues[time][maxGroup] =
--- a/giraffe/src/transforms/line.ts
+++ b/giraffe/src/transforms/line.ts
@@ -3,25 +3,50 @@ import {
   LineLayerSpec,
   LineData,
   LinePosition,
-  DomainLabel,
+  NumericColumnData,
+  CumulativeValuesByTime,
 } from '../types'
-import {FILL} from '../constants/columnKeys'
+import {FILL, VALUE} from '../constants/columnKeys'
 import {createGroupIDColumn, getNominalColorScale} from './'
 
-export const getPreviousDomainNextValue = (
-  lineData: LineData,
-  groupID: number,
-  domainLabel: DomainLabel
+export const mapCumulativeValuesToTimeRange = (
+  timesCol: NumericColumnData,
+  valuesCol: NumericColumnData,
+  fillCol: NumericColumnData
+): CumulativeValuesByTime => {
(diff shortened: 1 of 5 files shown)


giraffe/src/utils/tooltip.ts
```diff
@@ -159,7 +159,17 @@ export const getPointsTooltipData = (
       values: orderDataByValue(
         hoveredRowIndices,
         sortOrder,
-        hoveredRowIndices.map(i => Number(table.getColumn(FILL)[i]) + 1)
+        (() => {
```
hoorayimhelping: a couple of questions about this.

1. can you not just store the output of this function into an array variable and pass it to `orderDataByValue`? I feel like that would be clearer than an iife

1. this looks like a lot of looping and intermediate / ephemeral variable creation for code that gets called on render. 

we are mapping over `hoveredRowIndices` to create an array based on `groupColData`. i only looked through on the storybook, but `groupColData` was always sorted in ascending order, making this `sort` seem unnecessary. what are the cases where this `sort` is necessary? if it isn't necessary, we can turn an algorithm with complexity O(n + n + n + n) (I'd argue each successive loop and variable (due to garbage collection) creation matters when rendering) algorithm into an O(n) algorithm - basically cut 4 loops into 1. 

even if we need to `sort`, the last loop seems unnecessary.

```ts
  const sortedData = []
  hoveredRowIndices
    .map(hoveredRowIndex => groupColData[hoveredRowIndex])
    .sort()
    .forEach((groupId, index) => {
      lineCountByGroupId[groupId] = index + 1
      sortedData[index] = groupId
    })

  return sortedData
```

whatcha think? you obviously have more context than me.

TCL735: `hoveredRowIndices` is the problem here. It is an array of the indices of another array representing the data that is hovered. Currently, it seems like `hoveredRowIndices` is already sorted but there is **no guarantee that it will be sorted in the future**, if say there was a need to update the algorithm that finds the nearest hovered points. See https://github.com/influxdata/giraffe/blob/master/giraffe/src/utils/useHoverPointIndices.ts - there is no guarantee or comments that the output is sorted.

My change is:
1. Get all of the `groupIds` of the lines that are being hovered. (Imagine there are 50 lines on the screen, but only 5 lines are being hovered, because the other lines have no data at the hover point.)
2. Based on the hovered `groupIds`, we can determine how many total lines (cumulative) each line represents. `groupId` is sequential; higher `groupId` numbers represent potentially more total lines. If we sort the `groupIds` the lowest `groupId` will have just 1 line, and every following `groupId`  is sequentially one more line.
3. We want to map `groupId` to a count of total lines for that `groupId`, because **ultimately we want an array of the line counts**, but in the order of the hovered row indices, which may not be sorted. Thus, we use `hoveredRowIndices` to map the line counts.

giraffe/src/transforms/line.ts
```diff
@@ -3,25 +3,50 @@ import {
...
+  fillCol: NumericColumnData
+): CumulativeValuesByTime => {
+  const cumulativeValues: CumulativeValuesByTime = {}
+
+  timesCol.forEach((time, index) => {
+    if (!cumulativeValues[time]) {
+      cumulativeValues[time] = {}
+    }
+    const maxGroup = fillCol[index]
+    for (let i = 0; i < maxGroup; i += 1) {
```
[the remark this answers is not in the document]
TCL735: Good point here. I will add a short circuit so that the loop doesn't run needlessly.

giraffe/src/utils/tooltip.ts
TCL735: You are right about the IIFE. Will remove it for performance reasons.

PR #162, opened 2020-02-26
fix(ui): fixed tooltip bug that was causing giraffe to crash for scatterplots

Asalem1:
Closes https://app.zenhub.com/workspaces/applicationmonitoring-5e3b05772922e316b3a210a4/issues/influxdata/influxdb/16984

### Problem

Giraffe's sorting functionality was assuming that the lineData would exist, even though it was an option parameter

### Solution

Added defaults to the lineData prop and return the default hover index if the lineData is empty. This was simply the most straightforward solution, but it could be that we would need to implement a more dynamic solution that handles sorting for scatter plots specifically since some of those properties do not exist

### DIFF
--- a/giraffe/src/utils/fixtures/tooltip.ts
+++ b/giraffe/src/utils/fixtures/tooltip.ts
@@ -23,7 +23,7 @@ const getRandomNumber = (
 }
 
 export const COLUMN_KEY = 'cpu'
-export const POINT_KEY = 'cpu'
+export const POINT_KEY = 'disk'
 export const HOST_KEY = 'host'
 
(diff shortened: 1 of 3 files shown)


TCL735:
@Asalem1 thanks for catching this and the insightful analysis. You are right about the cause.

I suggest a slightly better fix. `lineData` should not be defaulted to an empty because that will result in an empty array for the desired sort order. Rather, we should default back to the order of the `hoveredRowIndices` if we do not have any lineData.

TCL735:
Requesting a third pair of eyes on this. @hoorayimhelping

TCL735:
Added unit test for scatterplot

PR #184, opened 2020-05-17
fix: add unique key to custom layer element

### LINKED ISSUE #183: Custom layer is missing unique key as part of a list of React elements
When rendering a custom layer on top of a graph, we create it as an element that is part of a list of elements in `<SizedPlot>`. Per React's best practices, a unique key is required for each element in a list.

Screen Shot 2020-05-17 at 3 32 40 PM

TCL735:
Closes #183 

- Following best practices, adds a unique key to the custom layer element in a list of React elements.
- Minor tweak to the `prettier` script to only fix files that we care about.

### DIFF
--- a/giraffe/package.json
+++ b/giraffe/package.json
@@ -9,8 +9,8 @@
     "build": "rm -rf dist && NODE_ENV=production rollup -c",
     "start": "rollup -cw",
     "lint": "eslint '{src,../stories/src}/**/*.{ts,tsx}'",
-    "prettier": "prettier --config ../.prettierrc.json --check '{.,../stories}/**/*.{ts,tsx}'",
-    "prettier:fix": "prettier --config ../.prettierrc.json --write '{.,../stories}/**/*.{ts,tsx}'",
+    "prettier": "prettier --config ../.prettierrc.json --check '{src,../stories/src}/**/*.{ts,tsx}'",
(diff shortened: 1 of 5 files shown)


giraffe/package.json
```diff
@@ -9,8 +9,8 @@
     "build": "rm -rf dist && NODE_ENV=production rollup -c",
     "start": "rollup -cw",
     "lint": "eslint '{src,../stories/src}/**/*.{ts,tsx}'",
-    "prettier": "prettier --config ../.prettierrc.json --check '{.,../stories}/**/*.{ts,tsx}'",
-    "prettier:fix": "prettier --config ../.prettierrc.json --write '{.,../stories}/**/*.{ts,tsx}'",
+    "prettier": "prettier --config ../.prettierrc.json --check '{src,../stories/src}/**/*.{ts,tsx}'",
```
hoorayimhelping, line 12: generally i think it's probably better to pull these kinds of tweaks out of PRs and get them reviewed ahead of the logic changes in that PR so that we can separate out housekeeping pieces like this and logical pieces like the `key` addition.

i've noticed that when these types of changes get merged together, and there's a problem in the logic, and the whole thing gets reverted, the housekeeping bits tend to not get re-added. i think  people tend to think of the housekeeping parts as done already and forget they got reverted. i've done this a number of times and seen teammates do it as well.

not going to block on this, but something to consider doing, going forward.

PR #213, opened 2020-06-17
fix: allow meridiem time format to be case insensitive

TCL735:
Closes https://github.com/influxdata/influxdb/issues/18458

Meridiem time should be case insensitive.

### DIFF
--- a/giraffe/src/utils/formatters.test.ts
+++ b/giraffe/src/utils/formatters.test.ts
@@ -18,6 +18,62 @@ describe('timeFormatter', () => {
     expect(nonUTCFormatter(d)).toEqual('2018-12-31  4:00:00 PM PST')
   })
 
+  test('handles ante meridiem and post meridiem', () => {
+    let d = new Date('2019-01-01T00:00Z')
+
+    let meridiemFormatter = timeFormatter({
+      timeZone: 'Europe/London',
+      format: 'HH:mm a',
+    })
+    expect(meridiemFormatter(d)).toEqual('12:00 AM')
+
+    meridiemFormatter = timeFormatter({
+      timeZone: 'Europe/London',
+      format: 'HH:mm A',
+    })
+    expect(meridiemFormatter(d)).toEqual('12:00 AM')
+
+    d = new Date('2019-01-01T13:00Z')
+
+    meridiemFormatter = timeFormatter({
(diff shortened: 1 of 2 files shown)


giraffe/src/utils/formatters.ts
```diff
@@ -143,7 +143,8 @@ export const timeFormatter = ({
     // and the total number of ticks consistent regardless of time frame
     hh: ({hour}) => (Number(hour) < 10 ? ` ${Number(hour)}` : String(hour)),
     HH: ({lhour}) => {
-      if (format && format.includes('a')) {
+      const hasMeridiem = / a/i
+      if (hasMeridiem.test(format)) {
```
Asalem1, line 147: interesting that this would be the case. Great find

giraffe/src/utils/formatters.test.ts
```diff
@@ -18,6 +18,62 @@ describe('timeFormatter', () => {
     expect(nonUTCFormatter(d)).toEqual('2018-12-31  4:00:00 PM PST')
   })
 
+  test('handles ante meridiem and post meridiem', () => {
+    let d = new Date('2019-01-01T00:00Z')
```
Asalem1, line 22: Is there any way we can update this test to reflect a different locale? These pass locally, but so did a few that Chris Henn wrote that ended up not working since they didn't mock out a locale. I'm not actually sure if this would work or if you'd need something like `moment` to do it:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

TCL735: Unfortunately, we can't. The jest tests are run inside node which does not support `Intl.DateTimeFormat.format`.

And I don't think changing the locale would help with the meridiem time in this issue. The bug was the 24 hour format was still showing alongside meridiem (am/pm). 12 hours were not being subtracted for PM hours. We have made the fix here with tests to ensure we subtract the 12 hours when using meridiem in both uppercase, lowercase, and even mixed case.

`Intl.DateTimeFormat` is for changing MM-DD-YYYY into their local equivalents. The existing formatter can handle DD-MM-YYYY and perhaps we should put in an option for that for international users. But that is out of scope for this issue.

PR #218, opened 2020-06-23
feat: allow Config to use csv from Flux directly

### LINKED ISSUE #186: API: auto-parse Flux results
Giraffe should handle raw Flux results in its API without requiring manual parsing and conversion to its internal `Table` data structure

TCL735:
Closes #186 

Updates Giraffe's API:

- allow directly passing in a comma separated value string from the result of a Flux query
- `fluxResponse` is the property name of the new interface
- maintain backwards compatibility with previous Giraffe API
- `table` still works and takes precedence over `fluxResponse` when both are present
- when both are missing, an empty table is created


giraffe/src/utils/formatters.ts
```diff
@@ -172,10 +172,15 @@ export const timeFormatter = ({
 
   let formatter
 
+  const getValidDate = (d: number): Date => {
+    const date = new Date(d)
```
hoorayimhelping, line 176:
```suggestion
  const getValidDate = (t: number): Date => {
    const date = new Date(t)
```
`t` represents a timestamp here, right? not a date object?

giraffe/README.md
```diff
@@ -70,19 +70,14 @@ A React-based visualization library powering the data visualizations in [InfluxD
 
 ## Example Using Flux [](#example-using-flux)
 
-When generating the table through a Flux result:
-
-- call the `fromFlux` utility function on the csv that is generated by Flux
-- get the table in the returned object from calling `fromFlux`
-
-Here is an example of turning a result in comma separate values (CSV) from Flux into a table and rendering it without optional properties:
+Use the string containing the comma separate values (CSV) from Flux as the `fluxResponse` property:
```
drdelambre, line 73: we should keep both usages in the documentation. since parsing large CSVs is a relatively expensive operation, we still need to be able to parse flux closer to the injection point (by the API) so that we can memoize it. i see this PR as making a better, fuller, API.. not a migration strategy towards a contract change

giraffe/src/utils/newTable.ts
```diff
@@ -100,3 +101,16 @@ class SimpleTable implements Table {
...
+
+export const newTableFromConfig = (config: Config): Table => {
+  if (!config) {
+    return newTable(0)
+  }
+  if (config.table) {
+    return config.table
+  }
+  if (config.fluxResponse) {
+    return fromFlux(config.fluxResponse).table
```
hoorayimhelping, line 113: i'm digging this logic, it's super clean!

when would `newTable` be called with a `config` that has a `table`? Could that be an indicator of repeated function calls?

giraffe/src/utils/formatters.ts
```diff
@@ -172,10 +172,15 @@ export const timeFormatter = ({
...
+    return date.getTime() === date.getTime() ? date : new Date()
+  }
+
   if (format) {
     // If a `format` string is passed, we simply use it
     formatter = (x: number) =>
```
hoorayimhelping, line 182: i know this isn't your code, but any chance we might rename `x` to something that hints at what this supposed to be? even `t` would be better cause it might represent `timestamp`

giraffe/src/types/index.ts
```diff
@@ -29,7 +29,8 @@ export interface Config {
   xScale?: string
   yScale?: string
 
-  table: Table
+  table?: Table
```
hoorayimhelping, line 32: i *think* i understand what's going on here, but i just want to make sure. [elsewhere in the pr](https://github.com/influxdata/giraffe/pull/218/files#diff-41549c804c9f3592a0ab56556e0096e9R113) there's an unsafe accesses to `table` - they assume that `table` will always exist. that is a slightly different data structure with a `table` property that is required, right?

giraffe/README.md
TCL735: Ok i will put the other usage back into the docs.

giraffe/src/utils/formatters.ts
TCL735: Correct.

And to give some context on why this was changed. When we have an empty table, negative Infinity and positive Infinity are passed in as the lower and upper bounds of the time range. Making a new `Date` with Infinity results in an invalid date object. Which in turn leads to a console error during rendering.

Hence, we use the current time as a placeholder for empty tables.

giraffe/src/utils/newTable.ts
TCL735: Normally, it `newTable` would not be called with a `config` that already has a `table`. This is to protect against edge cases and misuse.

giraffe/src/types/index.ts
TCL735: Yes. That unsafe access is on a different interface. The result of calling `fromFlux` is an object defined by the  `FromFluxResult` interface, which _always_ has a table property.

giraffe/src/utils/formatters.ts
TCL735: Good call on the variable naming. i updated it to use something more descriptive.

giraffe/src/utils/formatters.ts
TCL735: Good call. I have updated it to something more descriptive.

Issue #224, opened 2020-06-29
Add dataset with string-based values to storybooks


hoorayimhelping:
#### Background
you'll want to create a branch off of master with the relevant code to add to the dropdowns. you might need to prune some of your code to make this happen. if you're not sure how to do this with git, try googling something that describes your situation: `git add specific unstaged files to branch`

#### Acceptance Criteria
- [ ] Dataset with string-based values is available in the storybook dropdowns

PR #227, opened 2020-06-29
feat: added cpuString data set

reparker837:
Closes issue #224

This dataset is meant to be used only with the mosaic layer, so it will not work with any existing graphs.


stories/src/helpers.tsx
```diff
@@ -68,7 +69,7 @@ export const colorSchemeKnob = (initial?: string[]) =>
   )
 
 export const tableKnob = (initial?: Table) =>
-  select('Data', {CPU}, initial || CPU)
+  select('Data', {CPU, CPUString}, initial || CPU)
```
TCL735, line 72: Since this new knob is intended only for Mosaic Plots, why not create another knob instead of adding to the tableKnob.

Perhaps something like `mosaicDataKnob`?

ebb-tide: @TCL735 I wonder if it is useful to have a concept of the giraffe plot failing gracefully when it encounters data it doesn't expect? I was hoping to ask @reparker837 and @rbose22 to have the mosaic plot display an error instead of breaking when it finds numeric values, and so I wonder if it could be a goal to have other graphs display a similar error?

TCL735: Sure. That sounds good. We can leave this knob as it is changed, for now.

PR #240, opened 2020-07-08
Feat: added Y-axis to Mosaic Chart

### LINKED ISSUE #235: Add Y-axis to Mosaic Layer
Create a Y-axis with strings as labels for each individual series

rbose22:
Closes #235 

Co-authored-by: Rose Parker <reparker837@gmail.com>

Screen Shot 2020-07-08 at 4 00 25 PM

### DIFF
--- a/giraffe/src/components/Axes.tsx
+++ b/giraffe/src/components/Axes.tsx
@@ -102,12 +102,9 @@ export const drawAxes = ({
   context.textAlign = 'end'
   context.textBaseline = 'middle'
   const yDomainWidth = yDomain[1] - yDomain[0]
-  console.log('yticks', yTicks)
   if (yColumnType === 'string') {
-    // TODO: implement
     let y
     for (let i = 0; i < yColumnLabels.length; i++) {
-      // console.log('yColumnLabel', label)
       y = yScale(i) + margins.top - height / (yColumnLabels.length * 2)
 
       context.globalAlpha = 1
--- a/giraffe/src/components/Axes.tsx
+++ b/giraffe/src/components/Axes.tsx
@@ -103,6 +103,7 @@ export const drawAxes = ({
 
   if (yColumnType === 'string') {
     //TODO: implement
+    // work in progress
   } else {
     for (const yTick of yTicks) {
       let y
--- a/giraffe/src/components/Axes.tsx
+++ b/giraffe/src/components/Axes.tsx
@@ -101,12 +101,11 @@ export const drawAxes = ({
   context.textBaseline = 'middle'
   const yDomainWidth = yDomain[1] - yDomain[0]
   if (yColumnLabels) {
-    let y
+    context.globalAlpha = 1
+    context.fillStyle = tickFontColor
     for (let i = 0; i < yColumnLabels.length; i++) {
-      y = yScale(i) + margins.top - height / (yColumnLabels.length * 2)
+      const y = yScale(i) + margins.top - height / (yColumnLabels.length * 2)
 
-      context.globalAlpha = 1
-      context.fillStyle = tickFontColor
       context.fillText(
         yTickFormatter(yColumnLabels[i], {domainWidth: yDomainWidth}),
         margins.left - TICK_PADDING_RIGHT,
--- a/giraffe/src/components/Axes.tsx
+++ b/giraffe/src/components/Axes.tsx
@@ -3,7 +3,7 @@ import {useRef, useLayoutEffect, FunctionComponent, CSSProperties} from 'react'
 
 import {TICK_PADDING_RIGHT, TICK_PADDING_TOP} from '../constants'
 import {clearCanvas} from '../utils/clearCanvas'
-import {Margins, Scale, SizedConfig, Formatter, ColumnType} from '../types'
+import {Margins, Scale, SizedConfig, Formatter} from '../types'
 
 import {PlotEnv} from '../utils/PlotEnv'
 
@@ -30,7 +30,6 @@ interface DrawAxesOptions {
   xScale: Scale<number, number>
   yScale: Scale<number, number>
   config: SizedConfig
-  yColumnType?: ColumnType
   yColumnLabels?: string[]
 }
 
@@ -59,7 +58,6 @@ export const drawAxes = ({
     xAxisLabel,
     yAxisLabel,
   },
-  yColumnType,
   yColumnLabels,
 }: DrawAxesOptions) => {
   clearCanvas(canvas, width, height)
@@ -102,7 +100,7 @@ export const drawAxes = ({
   context.textAlign = 'end'
   context.textBaseline = 'middle'
   const yDomainWidth = yDomain[1] - yDomain[0]
-  if (yColumnType === 'string') {
+  if (yColumnLabels) {
     let y
     for (let i = 0; i < yColumnLabels.length; i++) {
       y = yScale(i) + margins.top - height / (yColumnLabels.length * 2)
@@ -229,7 +227,6 @@ export const Axes: FunctionComponent<Props> = ({env, style}) => {
     xScale,
     yScale,
     config,
-    yColumnType,
     yColumnLabels,
   } = env
 
@@ -248,7 +245,6 @@ export const Axes: FunctionComponent<Props> = ({env, style}) => {
       xScale,
       yScale,
       config,
-      yColumnType,
       yColumnLabels,
     })
   }, [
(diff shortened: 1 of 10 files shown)


giraffe/src/components/Axes.tsx
```diff
@@ -100,9 +100,19 @@ export const drawAxes = ({
   context.textAlign = 'end'
   context.textBaseline = 'middle'
   const yDomainWidth = yDomain[1] - yDomain[0]
+  if (yColumnLabels) {
+    let y
```
hoorayimhelping, line 104: I see why you added this, but it's not necessary here.

`let` and `const` are block scoped, which means they only live while the block they are declared in is in scope. This becomes relevant in `if/else` and `try/catch` statements:

```ts
if (someCondition) {
  const myVariable = 'true'
} else {
  const myVariable = 'false'
}
console.log(myVariable) // myVariable is not defined
```
`myVariable` won't exist at the `console.log` statement. So the solution is to do this:

```js
let myVariable
if (someCondition) {
  myVariable = 'true'
} else {
  myVariable = 'false'
}
console.log(myVariable) // hey it worked!
```

But this `if` block doesn't have any situation where something needs to use `y`  when it's out of scope. So all of this is to say that you can define and declare `y` on every iteration of the loop:

```ts
const y = yScale(i) + margins.top - height / (yColumnLabels.length * 2)
```

```suggestion
```

giraffe/src/components/Axes.tsx
```diff
@@ -100,9 +100,19 @@ export const drawAxes = ({
...
+    for (let i = 0; i < yColumnLabels.length; i++) {
+      y = yScale(i) + margins.top - height / (yColumnLabels.length * 2)
```
hoorayimhelping, line 106:
```suggestion
      const y = yScale(i) + margins.top - height / (yColumnLabels.length * 2)
```

giraffe/src/components/Axes.tsx
```diff
@@ -100,9 +100,19 @@ export const drawAxes = ({
...
 
-  if (yColumnType === 'string') {
-    //TODO: implement
+      context.globalAlpha = 1
+      context.fillStyle = tickFontColor
```
hoorayimhelping, line 109: I understand why you added these here, but they can be moved up above the loop.

canvas is a state machine. What that means is, you set the state, and it retains whatever state you set until you change the state by setting it to something else. Inside of a loop with no conditions (i.e. there's no `if` statements in our loop), you can move it to outside of the loop, since every item in the loop will use the same style.

giraffe/src/types/index.ts
```diff
@@ -58,8 +58,8 @@ export interface Config {
   legendCrosshairColor?: string
   legendColumns?: string[]
 
-  // The type of the y-axis column
-  yColumnType?: ColumnType
+  // The labels for the y-axis column
```
hoorayimhelping, line 61: ✂️  - i think this comment has become unnecessary because `yColumnLabels` is pretty darn self explanatory

giraffe/src/utils/PlotEnv.ts
```diff
@@ -149,22 +150,11 @@ export class PlotEnv {
...
-    return 'number'
-  }
-
-  public set yColumnType(columnType: ColumnType) {
-    this.config.yColumnType = columnType
+  public get yColumnLabels(): string[] {
+    return this.config.layers
+      .map((_, i) => this.getSpec(i))
+      .filter(spec => spec && spec.yColumnLabels)
+      .map(spec => spec.yColumnLabels)[0]
```
hoorayimhelping, line 157: let's clean this up a bit (the original algorithm was hacky and this is a bit more functional, but a bit unoptimized). I'm going to write out what I see in pseudocode and see if we can factor out the logic to be simpler and reflect what we're trying to do.

```
for every *config layer* in the *config layers collection*
  get the spec of this layer
  return the spec of this layer as an item in this array

for every *spec* in the *spec array* just created
  filter any item where spec is falsy or spec is truthy but its key *yColumnLabels* is falsy

for every *spec* in the *filtered spec array* just created
  return the *yColumnLabel* for this spec

get the 0th item of the array of *yColumnLabels*
  (which should be the first label we've encountered)
```
is that correct?

it seems like the intention is to get the first legitimate `yColumnLabel` from this collection and use that. Is that correct? 

If it's correct, can you think of a way to communicate that to someone reading this code without much context?

giraffe/src/utils/getMargins.ts
```diff
@@ -13,16 +13,33 @@ export const getMargins = (
...
-    yTicks.map(t => yTickFormatter(t))
-  )
+  let longestYTick
+  if (yColumnLabels) {
+    // for cases where they y-axis has strings as values
+    // find the longest label in the list
+    let maxVal = yColumnLabels[0]
+    let maxLength = maxVal.length
+    let i
+    for (i = 1; i < yColumnLabels.length; i++) {
```
hoorayimhelping, line 30: it looks like we want to use the function [maxBy](https://github.com/influxdata/giraffe/blob/master/giraffe/src/utils/extrema.ts#L21) similar to below. It seems like it's made for this. Of course it's hard to tell because the function is dense and looks like mathematical notation. After staring at it for a good 10 minutes, I *think* this is what it do:

The first argument is a function that gets run on every item of the array. That function should produce the value you're comparing. In this case, length of the label. The second argument is the array to get the max value for.

I **think** (haven't tested it) you want something like this:

```ts
  let longestYTick
  if (yColumnLabels) {
    longestYTick = maxBy(
      d => d.length,
      yColumnLabels
    )
  }
```

giraffe/src/utils/PlotEnv.ts
reparker837: here is the solution rashi & deniz came up with:

```
  public get yColumnLabels(): string[] {
    const currLayerSpec = this.config.layers.map((_, i) => this.getSpec(i))[0]

    if (currLayerSpec.type === 'mosaic') {
      return currLayerSpec.yColumnLabels
    }
    return []
  }
```

giraffe/src/components/Axes.tsx
```diff
@@ -100,9 +100,18 @@ export const drawAxes = ({
   context.textAlign = 'end'
   context.textBaseline = 'middle'
   const yDomainWidth = yDomain[1] - yDomain[0]
+  if (yColumnLabels) {
+    context.globalAlpha = 1
+    context.fillStyle = tickFontColor
+    for (let i = 0; i < yColumnLabels.length; i++) {
+      const y = yScale(i) + margins.top - height / (yColumnLabels.length * 2)
```
hoorayimhelping, line 107: might want to put a comment here saying it's possible to divide by zero here since `yColumnLabels.length` can never be `0` on this line. I noticed a potential divide by zero and had to work out that this is a safe division because of the loop above. A comment encodes that mental work into something a little less work for the readier :)

giraffe/src/utils/PlotEnv.ts
```diff
@@ -149,22 +150,13 @@ export class PlotEnv {
...
-    }
+  public get yColumnLabels(): string[] {
+    const currLayerSpec = this.config.layers.map((_, i) => this.getSpec(i))[0]
 
-    return 'number'
-  }
-
-  public set yColumnType(columnType: ColumnType) {
-    this.config.yColumnType = columnType
+    if (currLayerSpec.type === 'mosaic') {
```
hoorayimhelping, line 156: ahh, okay, it makes sense why this doesn't bother TypeScript. :+1:

Not relevant to shipping this code but from a software readability / design perspective, I'm less of a fan of this. It's ultimately fine, because TypeScript is guaranteeing this things safety, but it makes it less explicit and more work for the reader's comprehension.

This pattern forces the reader to look up the definition for mosaic to see that `yColumnLabels` will always be an array on that, vs having the code here explicitly say "if Array.isArray(currLayerSpec.yColumnLabels), return that array, otherwise return an empty array."

Ultimately minor thing in code review, but has a larger cost down the line for anyone reading this. Like I said, not something to worry about in this review, but something to think about from a larger picture perspective.

giraffe/src/utils/getMargins.ts
```diff
@@ -13,16 +13,24 @@ export const getMargins = (
   yAxisLabel: string,
   yTicks: number[],
   yTickFormatter: (tick: number) => string,
-  tickFont: string
+  tickFont: string,
+  yColumnLabels: string[]
```
ebb-tide, line 17: Why not use yTicks to store yColumnLabels and have yTicks be of type `number[]` for non-mosaic layers, and `string[]` for mosaic layers?

ebb-tide: and then yTickFormatter could be the identity function and be defined as type : `(tick: number| string)=> string`

PR #243, opened 2020-07-10
Feat: added Y-axis to Mosaic Chart 

### LINKED ISSUE #235: Add Y-axis to Mosaic Layer

rbose22:
Closes #235 

Co-authored-by: Rose Parker reparker837@gmail.com

This PR adds a Y-axis with string values to the mosaic chart. It achieves the same goal that the previous PR did, but we used yTicks in this version instead of creating a new variable. This change allowed us to use more of the existing code, rather than creating specific functions that would only be used in the mosaic graph case.

### DIFF
--- a/giraffe/src/components/Axes.tsx
+++ b/giraffe/src/components/Axes.tsx
@@ -102,28 +102,7 @@ export const drawAxes = ({
   let count = 0
   for (const yTick of yTicks) {
     let y
-    if (typeof yTick === 'number') {
-      y = yScale(yTick) + margins.top
-      if (
-        Math.abs(y - margins.top) > GRID_LINE_MIN_DIST &&
-        Math.abs(y - (height - margins.bottom)) > GRID_LINE_MIN_DIST
-      ) {
-        context.strokeStyle = gridColor
-        context.globalAlpha = gridOpacity
-        context.beginPath()
-        context.moveTo(margins.left, y)
-        context.lineTo(width - margins.right, y)
-        context.stroke()
-      }
-
-      context.globalAlpha = 1
-      context.fillStyle = tickFontColor
-      context.fillText(
-        yTickFormatter(yTick, {domainWidth: yDomainWidth}),
-        margins.left - TICK_PADDING_RIGHT,
-        y
-      )
-    } else if (typeof yTick === 'string') {
+    if (typeof yTick === 'string') {
       y = yScale(count) + margins.top - height / (yTicks.length * 2)
       context.globalAlpha = 1
       context.fillStyle = tickFontColor
@@ -135,15 +114,7 @@ export const drawAxes = ({
       )
       count += 1
     } else {
-      y = yTick
-
-      context.globalAlpha = 1
-      context.fillStyle = tickFontColor
-      context.fillText(
-        yTickFormatter(yTick, {domainWidth: yDomainWidth}),
-        margins.left - TICK_PADDING_RIGHT,
-        y
-      )
+      y = yScale(yTick) + margins.top
       if (
         Math.abs(y - margins.top) > GRID_LINE_MIN_DIST &&
         Math.abs(y - (height - margins.bottom)) > GRID_LINE_MIN_DIST
--- a/giraffe/src/components/Axes.tsx
+++ b/giraffe/src/components/Axes.tsx
@@ -58,8 +58,7 @@ export const drawAxes = ({
     xAxisLabel,
     yAxisLabel,
   },
-}: //yColumnType,
-DrawAxesOptions) => {
+}: DrawAxesOptions) => {
   clearCanvas(canvas, width, height)
 
   const context = canvas.getContext('2d')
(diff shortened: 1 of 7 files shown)


giraffe/src/components/Axes.tsx
```diff
@@ -100,55 +99,70 @@ export const drawAxes = ({
...
-          context.lineTo(width - margins.right, y)
-          context.stroke()
-        }
+      context.fillText(
+        yTickFormatter(yTick, {domainWidth: yDomainWidth}),
+        margins.left - TICK_PADDING_RIGHT,
+        y
+      )
+      count += 1
+    } else {
```
ebb-tide, line 137: could yTick be something other than number or string?

giraffe/src/types/index.ts
```diff
@@ -255,6 +255,7 @@ export interface MosaicLayerConfig {
   strokeOpacity?: number
   fillOpacity?: number
   yColumnType: ColumnType
+  yTicks: string[]
```
ebb-tide, line 258: Amazing. now typescript should know if config is of type mosaic, yTicks must be string.

giraffe/src/utils/getMargins.ts
```diff
@@ -11,18 +11,26 @@ export const getMargins = (
...
-    d => d.length,
-    yTicks.map(t => yTickFormatter(t))
-  )
+  let longestYTick
+  if (typeof yTicks[0] === 'string') {
+    longestYTick = maxBy(
+      d => d.length,
+      (yTicks as string[]).map(t => yTickFormatter(t))
+    )
+  } else {
```
ebb-tide, line 28: do you still need this else? they look pretty similar.. except for typescript stuff.

giraffe/src/utils/PlotEnv.ts
```diff
@@ -104,11 +103,18 @@ export class PlotEnv {
...
 
-  public get yTicks(): number[] {
+  public get yTicks(): number[] | string[] {
+    const currLayerSpec = this.config.layers.map((_, i) => this.getSpec(i))[0]
+    if (
+      currLayerSpec &&
+      currLayerSpec.type === 'mosaic' &&
+      Array.isArray(currLayerSpec.yTicks)
+    ) {
+      return currLayerSpec.yTicks
```
ebb-tide, line 113: hrrm.. why are other yTicks directly on config and ours is on currLayerSpec? just confused.

giraffe/src/components/Axes.tsx
reparker837: I don't believe so, since time is handled as a number

hoorayimhelping: so this `else` existed to handle when there was a string appended to the number, like a number with a label, right?

does the `else if (typeof yTick === 'string')` conditional cover all the logic down here? i.e. have you tested it with a number that has a label? it looks like this `else`, which the number with label would hit, isn't doing the same things as the `else if (typeof yTick === 'string')` above and so I wouldn't expect the storybook to act the same.

if `yTick` could only be a number or a string, we need to remove one of these `else` conditionals.

giraffe/src/utils/getMargins.ts
```diff
@@ -11,18 +11,26 @@ export const getMargins = (
   showAxes: boolean,
   xAxisLabel: string,
   yAxisLabel: string,
-  yTicks: number[],
-  yTickFormatter: (tick: number) => string,
+  yTicks: string[] | number[],
```
ebb-tide, line 14:
```suggestion
  yTicks: Array<string | number>,
```

giraffe/src/utils/getMargins.ts
```diff
@@ -11,18 +11,26 @@ export const getMargins = (
...
+    longestYTick = maxBy(
+      d => d.length,
+      (yTicks as string[]).map(t => yTickFormatter(t))
+    )
+  } else {
+    longestYTick = maxBy(
+      d => d.length,
+      (yTicks as number[]).map(t => yTickFormatter(t))
+    )
+  }
```
ebb-tide, line 33: const formattedYTicks = yTicks.map(t => yTickFormatter(t))
  const longestYTick = maxBy(d => d.length, formattedYTicks)

ebb-tide: This should work!

rbose22: I'm not sure if I'm doing something wrong or missing some code elsewhere, but when I make the changes you suggested above I get 2 errors. It says map is not callable because the signatures are not compatible with each other (line 22), and it says property length does not exist on type 'unknown' (line 23). Do you have any ideas on why TS might be complaining on my version but not on yours?

ebb-tide: you need to also make the change I suggested above for it to work.  ` yTicks: string[] | number[]` to  `  yTicks: Array<string | number>`

ebb-tide: let me know if it still doesn't work

rbose22: Never mind. Sorry missed the above suggestion about changing the type. It works!

PR #244, opened 2020-07-11
feat: add RawFluxDataTable visualization

### LINKED ISSUE #171: Move raw data view into giraffe

TCL735:
Closes #171 

Raw Flux Data Table view
- needed CSS Modules to support styling
- borrowed (aka copy/paste) `DapperScrollbars` from clockface rather than importing it (to keep build sizes small)


TCL735:
Unresolved issues:

- The generated .d.ts files for the styles have semicolons in them, which is causing a lint error in the CI
- Removing the semicolons does not work, because every time webpack builds, the semicolons are re-inserted
- The `DapperScrollbar` horizontal scroll is unresponsive and hard to use. The onScroll handler is problematic.

giraffe/src/types/index.ts
```diff
@@ -163,6 +167,14 @@ export interface CustomLayerRenderProps {
   columnFormatter: (colKey: string) => (x: any) => string
 }
 
+export interface RawFluxDataTableLayerConfig {
+  type: 'flux data table' // do not refactor or restrict to LayerTypes.RawFluxDataTable
```
hoorayimhelping, line 171: out of curiosity, why not? it's good to remember why we do things so we don't end up in a [monkeys on a ladder](https://www.youtube.com/watch?v=y-PvBo75PDo) situation

giraffe/src/utils/fluxParsing.test.ts
```diff
@@ -0,0 +1,142 @@
...
+  })
+
+  describe('group key', () => {
+    test('parses the group key properly', () => {
+      const actual = parseResponse(MULTI_SCHEMA_RESPONSE)[0].groupKey
+      const expected = {
+        _field: 'usage_guest',
+        _measurement: 'cpu',
+        cpu: 'cpu-total',
+        host: 'WattsInfluxDB',
```
hoorayimhelping, line 129: immortalizing watts eh?

giraffe/src/utils/fluxParsing.ts
```diff
@@ -0,0 +1,149 @@
...
+export const parseTables = (responseChunk: string): FluxTable[] => {
+  const lines = responseChunk.split('\n')
+  const annotationLines: string = lines
+    .filter(line => line.startsWith('#'))
+    .join('\n')
+    .trim()
+  const nonAnnotationLines: string = lines
+    .filter(line => !line.startsWith('#'))
+    .join('\n')
+    .trim()
```
hoorayimhelping, line 74:
```suggestion
```

giraffe/src/utils/fluxParsing.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  if (!annotationLines) {
+    throw new Error('Unable to extract annotation data')
+  }
+
```
hoorayimhelping, line 79:
```suggestion
  const nonAnnotationLines: string = lines
    .filter(line => !line.startsWith('#'))
    .join('\n')
    .trim()
```

hoorayimhelping: suggestion: see if we can early return before creating a new variable

giraffe/src/utils/groupBy.ts
```diff
@@ -0,0 +1,14 @@
+type Iteratee = (value: any) => string
+
+export const groupBy = (data: Array<any>, iteratee: Iteratee) => {
+  const result = {}
+
+  data.forEach(value => {
+    const index = iteratee(value)
+    if (!result[index]) {
```
hoorayimhelping, line 8:
```suggestion
    if (!Array.isArray(result[index])) {
```

giraffe/src/utils/rawFluxDataTable.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (i !== 0) {
+      // Seperate each chunk by an empty line, just like in the unparsed CSV
+      data.push([])
+    }
+
+    for (let j = 0; j < parsedChunks[i].length; j++) {
+      // Danger zone! Since the contents of each chunk are potentially quite
+      // large, the contents need to be concated using a loop rather than with
+      // `concat`, a splat or similar. Otherwise we see a "Maximum call size
+      // exceeded" error for large CSVs
```
hoorayimhelping, line 26: :+1: good comment

giraffe/src/types/index.ts
TCL735: This is for storybook and code examples.

Storybook examples are harder to work with when TypeScript restricts them to an enum. It causes storybook to import the enum, which by itself, is not a problem. But then this will also extend out to any other example code, including in the README.

The examples in the README become convoluted, since we are now referring to something that TypeScript requires rather than just a simple string. The README code is more approachable if we allow a string "flux data table" rather than require new users to already have a solid grasp of TypeScript.

giraffe/src/utils/fluxParsing.ts
TCL735: Am I missing something? This looks the same.

giraffe/src/utils/fluxParsing.ts
TCL735: Done.

giraffe/src/utils/groupBy.ts
TCL735: Done.

giraffe/src/components/RawFluxDataGrid.tsx
```diff
@@ -0,0 +1,133 @@
...
+  height: number
+  scrollLeft: number
+  scrollTop: number
+}
+
+interface State {
+  headerRows: number[]
+}
+
+const getCellData = (data: string[][], row, column) => {
```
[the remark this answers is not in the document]
TCL735: String. Done.

giraffe/src/components/DapperScrollbars.tsx
```diff
@@ -0,0 +1,204 @@
...
+      id={id}
+      download={null}
+      inlist={null}
+    >
+      {children}
+    </Scrollbar>
+  )
+}
+
+DapperScrollbars.displayName = 'DapperScrollbars'
```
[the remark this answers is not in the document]
TCL735: The issue here is the displayName being the same string as used elsewhere. I will change the displayName to **'GiraffeDapperScrollbars'**

giraffe/src/components/RawFluxDataGrid.scss
```diff
@@ -0,0 +1,40 @@
...
+  &:hover {
+    z-index: 2;
+
+    .raw-flux-data-table--cell-bg {
+      border: 2px solid #00a3ff;
+      background: #383846;
+      width: auto;
+    }
+  }
+}
```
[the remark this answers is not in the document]
TCL735: i updated it with more alphabetizing--in a way that groups related CSS properties together and alphabetizes them.

giraffe/src/utils/fluxParsing.test.ts
TCL735: There must always be **watts**.

Issue #236, opened 2020-07-07
Add Tooltip to Mosaic Layer


rbose22:
- Decide what information should appear on the tooltip legend 
- What the display behavior should be
- Implement the tooltip legend

PR #246, opened 2020-07-14
Feat: added mosaic tooltip legend

reparker837:
Closes issue #236 

Added a tooltip legend that displays the time range, value and series of the data interval (color block) you're currently hovering over.

Note: legend font currently looks different from other graphs (as we haven't set it with knobs as the others have)

Co-authored-by: Rashi Bose <rashibose@gmail.com>


giraffe/src/utils/mosaicTooltip.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+  const xMinData = boxTable.getColumn(X_MIN, 'number')
+  const xMaxData = boxTable.getColumn(X_MAX, 'number')
+  const dataX = xScale.invert(hoverX)
+  const dataY = yScale.invert(hoverY)
+  // Find all boxes whose x extent contain the mouse x position
+  const xIndices = range(0, xMinData.length).filter(
+    i => xMinData[i] <= dataX && xMaxData[i] > dataX
+  )
+
```
hoorayimhelping, line 32: first things first. `xIndicies` is referenced directly in the return statement of this function:

```ts
  if (xIndices) {
    xyIndices = [xIndices[finalXIndex]]
  } else {
    xyIndices = []
  }

   return xyIndices
```

if this comes back as falsy, we can just early return from this function right now without any more looping.

```suggestion
if (!xIndices) {
  return []
}
```

now, any code below this `if` statement (which i've heard referred to as guard statements) is considered happy path - it can be assumed to work, since we checked an exit condition early. this pattern (where we check if there are any conditions that will cause our algorithm to fail, and if so, bail) is called an early return.

this early return can drastically change the looping code below, because now loops can return early as well, rather than maintaining state in a variable, and handling a return based on that (see below)

giraffe/src/utils/mosaicTooltip.ts
```diff
@@ -0,0 +1,115 @@
...
+  let finalXIndex
+  for (const index of yValMap.keys()) {
+    if (
+      yValMap.get(index)[0] < yScale(dataY) &&
+      yValMap.get(index)[1] >= yScale(dataY)
+    ) {
+      finalXIndex = index
+      break
+    }
+  }
```
hoorayimhelping, line 50: i'm looking at both of these loops together. a [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) is an object that maintains looping order based on sorting order. Since the items in `yValMap` are being inserted in the order they're looped over, and the main reason to use a `Map` is to maintain order when looping, I think we can combine the functionality of the loops and only loop over these values once.

```ts
for (let i = 0; i < yDomain; i++) {
  const yMin = yScale(i + 1)
  const yMax = yScale(i)

  if (yMin < yScale(dataY) && yMax >= yScale(dataY) {
    return [xIndicies[index]] // if we return early above when xIndicies is falsy this can be changed to a return
  }
}
 // handles the case where the loop didn't return early.
return []
```

giraffe/src/utils/mosaicTooltip.ts
```diff
@@ -0,0 +1,115 @@
...
+  const colors = hoveredBoxRows.map(i =>
+    fillScale((valCol[i] as unknown) as number)
+  )
+
+  const xTooltipColumn: TooltipColumn = {
+    key: xColKey,
+    name: inputTable.getColumnName(xColKey),
+    type: inputTable.getColumnType(xColKey),
+    colors,
+    values: hoveredBoxRows.map(i =>
```
hoorayimhelping, line 87: `i` usually represents an integer counter. something that gets incremented in a loop. I'd like to see this changed to something like `h` or `box` or `row` or `r` or `hbr`. But not `i`, `j`, or `k` cause those are almost universally integer counters.

giraffe/src/utils/mosaicTooltip.ts
```diff
@@ -0,0 +1,115 @@
...
+      getRangeLabel(xMinCol[i], xMaxCol[i], xFormatter)
+    ),
+  }
+
+  const yTooltipColumn: TooltipColumn = {
+    key: yColKey,
+    name: inputTable.getColumnName(yColKey),
+    type: inputTable.getColumnType(yColKey),
+    colors,
+    values: hoveredBoxRows.map(i => yFormatter(yCol[i])),
```
hoorayimhelping, line 97: same bit about `i`

giraffe/src/utils/mosaicTooltip.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  const groupTooltipColumns = fillGroupMap.columnKeys.map(key => ({
+    key,
+    name: inputTable.getColumnName(key),
+    type: inputTable.getColumnType(key),
+    colors,
+    values: hoveredBoxRows.map(i => valFormatter(valCol[i])),
```
hoorayimhelping, line 105: same bit about `i`

giraffe/src/utils/drawMosaic.ts
```diff
@@ -60,9 +54,13 @@ export const drawMosaic = ({
     const width = xScale(xMaxCol[i]) - x - strokePadding
     const height = yScale(yValMap.size + 1)
 
-    const colorVal = colorMap.get(valueCol[i])
+    let fill = fillScale((valueCol[i] as unknown) as number)
```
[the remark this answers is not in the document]
reparker837: You're right! I was able to change the code to this:
```
const valCol = (table.getColumn(FILL, 'string') as unknown) as Scale<number, number>
let fill = fillScale(valueCol[i])
```

giraffe/src/utils/mosaicTooltip.ts
```diff
@@ -0,0 +1,115 @@
...
+  Scale,
+  TooltipData,
+  TooltipColumn,
+  ColumnGroupMap,
+} from '../types'
+
+export const findHoveredBoxes = (
+  boxTable: Table,
+  hoverX: number | null,
+  hoverY: number | null,
```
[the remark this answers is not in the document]
reparker837: Looks like they're set explicitly to null  

```
const hoverX = dragEvent ? null : hoverEvent.x
const hoverY = dragEvent ? null : hoverEvent.y
```

giraffe/src/utils/mosaicTooltip.ts
reparker837: I think that change makes sense. However, the other tooltip files use the variable i for mapping as well. Is it better to change the variable name in this file for clarity or leave it the same as the others for consistency?

giraffe/src/components/MosaicLayer.tsx
```diff
@@ -40,6 +61,22 @@ export const MosaicLayer: FunctionComponent<Props> = ({
     Object.values(drawMosaicOptions)
   )
 
+  let tooltipData: TooltipData = []
+  if (hoveredRowIndices.length > 0) {
+    console.log('case entered')
+    console.log('indices', hoveredRowIndices)
```
ebb-tide: why we have console.logs? ;)

giraffe/src/utils/mosaicTooltip.ts
```diff
@@ -0,0 +1,134 @@
...
+      return [xIndices[i]] // isn't there some value to referring to this as xyIndices for clarity?
+    }
+  }
+  // handles the case where the loop didn't return early.
+  return []
+
+  //const yValMap = new Map()
+  //if series isn't in map yet, add it
+
+  // for (let i = 0; i < yDomain[1]; i++) {
```
ebb-tide: why so much comments here?

giraffe/src/utils/rectTooltip.ts
```diff
@@ -39,11 +39,9 @@ export const findHoveredRects = (
   } else if (binDimension === 'x') {
     return xIndices
   }
-
```
ebb-tide, line 42: in general I'd avoid including these changes that don't do anything. it adds you as the last person who modified this file in git. Make sure you are using `git add --patch` to stage your changes one by one.  :)

giraffe/src/utils/PlotEnv.ts
```diff
@@ -287,7 +287,7 @@ export class PlotEnv {
           layerConfig.y,
           this.config.xDomain,
           layerConfig.fill,
-          layerConfig.symbol,
+          //layerConfig.symbol,
```
ebb-tide, line 290: I'm assuming you'll remove these? if not let me know why not plz. :)

giraffe/src/transforms/mosaic.ts
```diff
@@ -11,33 +11,33 @@ export const mosaicTransform = (
   yColumnKey: string,
   xDomain: number[],
   fillColKeys: string[],
-  symbolColKeys: string[],
+  //symbolColKeys: string[],
```
hoorayimhelping, line 14: ✂️

giraffe/src/transforms/mosaic.ts
```diff
@@ -11,33 +11,33 @@ export const mosaicTransform = (
...
   )
 
-  const [symbolColumn, symbolColumnMap] = createGroupIDColumn(
-    inputTable,
-    symbolColKeys
-  )
+  // const [symbolColumn, symbolColumnMap] = createGroupIDColumn(
+  //   inputTable,
+  //   symbolColKeys
+  // )
```
hoorayimhelping, line 25: ✂️

giraffe/src/transforms/mosaic.ts
```diff
@@ -11,33 +11,33 @@ export const mosaicTransform = (
...
 
   // break up into itervals while adding to table
   const xMinData = []
   const xMaxData = []
   const fillData = []
   const seriesData = []
-  const symbolData = []
+  //const symbolData = []
```
hoorayimhelping, line 32: ✂️

giraffe/src/transforms/mosaic.ts
```diff
@@ -11,33 +11,33 @@ export const mosaicTransform = (
...
   let prevValue = ''
   let tableLength = 0
   let prevSeries = inputTable.getColumn(yColumnKey, 'string')[0]
 
   // find all series in the data set
   const valueStrings = [inputTable.getColumn(yColumnKey, 'string')[0]]
 
-  const valueType1 = symbolColumnMap.columnKeys[0]
+  //const valueType1 = symbolColumnMap.columnKeys[0]
```
hoorayimhelping, line 40: ✂️

giraffe/src/transforms/mosaic.ts
```diff
@@ -46,7 +46,7 @@ export const mosaicTransform = (
   xMinData.push(inputTable.getColumn(xColumnKey, 'number')[0])
   fillData.push(fillColumnMap.mappings[fillColumn[0]][valueType2])
   seriesData.push(inputTable.getColumn(yColumnKey, 'string')[0])
-  symbolData.push(symbolColumnMap.mappings[symbolColumn[0]][valueType1])
+  //symbolData.push(symbolColumnMap.mappings[symbolColumn[0]][valueType1])
```
hoorayimhelping, line 49: ✂️

giraffe/src/transforms/mosaic.ts
```diff
@@ -57,15 +57,15 @@ export const mosaicTransform = (
       xMinData.push(inputTable.getColumn(xColumnKey, 'number')[i])
       fillData.push(fillColumnMap.mappings[fillColumn[i]][valueType2])
       seriesData.push(inputTable.getColumn(yColumnKey, 'string')[i])
-      symbolData.push(symbolColumnMap.mappings[symbolColumn[i]][valueType1])
+      //symbolData.push(symbolColumnMap.mappings[symbolColumn[i]][valueType1])
```
hoorayimhelping, line 60: ✂️

giraffe/src/transforms/mosaic.ts
```diff
@@ -57,15 +57,15 @@ export const mosaicTransform = (
...
       prevSeries = inputTable.getColumn(yColumnKey, 'string')[i]
       tableLength += 1
     } else if (fillColumnMap.mappings[fillColumn[i]][valueType2] != prevValue) {
       xMaxData.push(inputTable.getColumn(xColumnKey, 'number')[i])
       xMinData.push(inputTable.getColumn(xColumnKey, 'number')[i])
       fillData.push(fillColumnMap.mappings[fillColumn[i]][valueType2])
       seriesData.push(inputTable.getColumn(yColumnKey, 'string')[i])
-      symbolData.push(symbolColumnMap.mappings[symbolColumn[i]][valueType1])
+      //symbolData.push(symbolColumnMap.mappings[symbolColumn[i]][valueType1])
```
hoorayimhelping, line 68: ✂️

giraffe/src/transforms/mosaic.ts
```diff
@@ -90,7 +90,7 @@ export const mosaicTransform = (
     .addColumn(X_MAX, 'number', xMaxData) //endTimes
     .addColumn(FILL, 'string', fillData) //values
     .addColumn(SERIES, 'string', seriesData) //cpus
-    .addColumn(SYMBOL, 'string', symbolData) //hosts
+  //.addColumn(SYMBOL, 'string', symbolData) //hosts
```
hoorayimhelping, line 93: ✂️

giraffe/src/utils/PlotEnv.ts
```diff
@@ -293,7 +293,7 @@ export class PlotEnv {
...
```
hoorayimhelping, line 296: ✂️

giraffe/src/utils/mosaicTooltip.ts
```diff
@@ -0,0 +1,120 @@
...
+  const xMaxData = boxTable.getColumn(X_MAX, 'number')
+  const dataX = xScale.invert(hoverX)
+  const dataY = yScale.invert(hoverY)
+  // Find all boxes whose x extent contain the mouse x position
+  const xIndices = range(0, xMinData.length).filter(
+    i => xMinData[i] <= dataX && xMaxData[i] > dataX
+  )
+
+  if (hoverDimension === 'x') {
+    return xIndices.reverse()
```
hoorayimhelping, line 35: will this always be an array?

do we need to add in the

```ts
if (!xIndicies) {
  return []
}
```

bit?

giraffe/src/utils/mosaicTooltip.ts
hoorayimhelping: great question. Either or works. Since there is technical cruft in the form of a poorly named variable elsewhere, we'll want to add a ticket to address that - It's out of the scope of this issue to fix that poorly named variable, but we want to address it, hence the ticket. Since we'll need to add a ticket to address that, and we'll have to do that later, it doesn't really matter all that much if we change this variable now, or later. If we change it now, cool. If we don't change it now, we'll have to also note on the ticket that we need to change the variable name here.

giraffe/src/transforms/mosaic.ts
```diff
@@ -11,33 +11,25 @@ export const mosaicTransform = (
...
   fillColKeys: string[],
-  symbolColKeys: string[],
   colors: string[]
 ): MosaicLayerSpec => {
   const [fillColumn, fillColumnMap] = createGroupIDColumn(
     inputTable,
     fillColKeys
   )
 
-  const [symbolColumn, symbolColumnMap] = createGroupIDColumn(
```
ebb-tide, line 22: yayy!

giraffe/src/utils/drawMosaic.ts
```diff
@@ -59,10 +53,13 @@ export const drawMosaic = ({
...
     const width = xScale(xMaxCol[i]) - x - strokePadding
     const height = yScale(yValMap.size + 1)
+    let fill = fillScale((valueCol[i] as unknown) as number)
 
-    const colorVal = colorMap.get(valueCol[i])
-
-    const fill = fillScale(colorVal)
+    if (hoveredRowIndices && hoveredRowIndices.includes(i)) {
+      fill = color(fill)
+        .brighter(1)
```
ebb-tide, line 60: ooh cool

PR #252, opened 2020-07-31
fix: handling overlapping data in mosaic transform

reparker837:
Cleaned up code and added a check to see if data is being overwritten. If so, the mosaic transform function now stops collecting data. 

Co-authored-by: rbose22 <rashibose@gmail.com>


giraffe/src/transforms/mosaic.ts
```diff
@@ -18,54 +18,61 @@ export const mosaicTransform = (
     fillColKeys
   )
 
+  const xInputCol = inputTable.getColumn(xColumnKey, 'number')
+  const yInputCol = inputTable.getColumn(yColumnKey, 'string')
```
hoorayimhelping, line 22: :+1:

PR #257, opened 2020-08-07
feat: implement Band Chart

### LINKED ISSUE #228: Visualization: Band Chart
**Given** I am constructing a query with numeric values
**When** I select Band Chart in the visualization dropdown in the Data Explorer
**And** Aggregate Functions `MIN`, `MAX` and `MEAN` are selected
**Then** `MIN` and `MAX` appear as shaded range as a band around the `MEAN` line

Design Notes:
As part of the Auto Aggregation epic, queries will be aggregated to improve performance. Single Series will get `Min`, `Max` and `Mean`

Design working doc
https://www.figma.com/file/naxHHmkcY0Tc6pfbgzdq9T/Auto-Aggregation?node-id=250%3A110

Other examples:
A
B

TCL735:
Closes #228 

Implements Band Chart:
Screen Shot 2020-08-06 at 9 51 33 PM


giraffe/src/components/BandTooltip.tsx
```diff
@@ -0,0 +1,89 @@
...
+    legendColumns: columnsWhitelist,
+  } = config
+
+  const columns = columnsWhitelist
+    ? data.filter(column => columnsWhitelist.includes(column.key))
+    : data
+
+  return createPortal(
+    <div
+      className="giraffe-tooltip"
```
ebb-tide, line 31: test review comment will delete

giraffe/src/transforms/band.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  const table = inputTable.addColumn(FILL, 'number', fillColumn)
+  const xCol = table.getColumn(xColumnKey, 'number')
+  const yCol = table.getColumn(yColumnKey, 'number')
+  const fillScale = getNominalColorScale(fillColumnMap, colors)
+  const bandFillColors = []
+  const lineData: LineData = {}
+  let stackedValuesByTime: CumulativeValuesByTime = {}
+
+  if (position === 'stacked') {
```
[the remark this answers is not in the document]
TCL735: nope, will remove

giraffe/src/components/BandLayer.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const hasHoverData = hoverRowIndices && hoverRowIndices.length > 0
+
+  return (
+    <>
+      <canvas
+        className="giraffe-layer giraffe-layer-band-chart"
+        ref={canvasRef}
+        style={{
+          position: 'absolute',
+          opacity: hoverDimension === 'xy' && hasHoverData ? 0.4 : 1,
```
reparker837, line 88: I'm not sure exactly what the issue is in the code, but when I change the hover dimension to 'xy' and try to hover over the graph, the graph disappears. Will changing the hover dimension be an option in the UI or is this just a giraffe issue?
ezgif-6-ea5b3ef913cd

giraffe/src/transforms/band.ts
```diff
@@ -0,0 +1,251 @@
...
+      } else {
+        bandIndices.rowIndices.push(index)
+      }
+    })
+  }
+
+  return bandIndices
+}
+
+export const getBandLineIndices = (fillColumnMap: ColumnGroupMap): number[] => {
```
reparker837, line 87: I'm not sure if this is because the lines are based off the fillColumn, but when you unselect the fill: cpu knob, all of the lines connect. Is this an expected behavior or an issue?
ezgif-6-be883319c754

giraffe/src/components/BandLayer.tsx
TCL735: Good catch. This is band chart specific. I need to update the hover points when the dimension is 'xy'

giraffe/src/transforms/band.ts
TCL735: Yes, this is correct in the sense that there is no way Flux will return a CSV without the cpu. This is an issue with the storybook knobs allowing a customization that would never happen. I'll adjust or possibly remove them.

giraffe/src/transforms/band.ts
```diff
@@ -0,0 +1,210 @@
...
+  const cumulativeValues: CumulativeValuesByTime = {}
+
+  timesCol.forEach((time, index) => {
+    if (!cumulativeValues[time]) {
+      cumulativeValues[time] = {}
+    }
+    const maxGroup = fillCol[index]
+
+    if (!cumulativeValues[time][maxGroup - 1]) {
+      for (let i = 0; i < maxGroup; i += 1) {
```
hoorayimhelping, line 125: was curious about `i += 1` vs `i++`. i've noticed them used interchangeably in giraffe and even this PR. is there a preference or reason to use one or the other?

giraffe/src/transforms/band.ts
```diff
@@ -0,0 +1,210 @@
...
+  const xCol = table.getColumn(xColumnKey, 'number')
+  const yCol = table.getColumn(yColumnKey, 'number')
+  const fillScale = getNominalColorScale(fillColumnMap, colors)
+  const bandFillColors = []
+  const lineData: LineData = {}
+
+  let xMin = Infinity
+  let xMax = -Infinity
+  let yMin = Infinity
+  let yMax = -Infinity
```
hoorayimhelping, line 161: these seem backwards - i'd expect `xMin` to be `-Infinity` and `xMax` to `Infinity`

giraffe/src/transforms/band.ts
```diff
@@ -0,0 +1,210 @@
...
+      xMax = x
+    }
+
+    if (y < yMin) {
+      yMin = y
+    }
+
+    if (y > yMax) {
+      yMax = y
+    }
```
hoorayimhelping, line 190: alternatively, this can be expressed in 4 lines:

```ts
xMin = Math.min(x, xMin)
xMax = Math.max(x, xMax)
yMin = Math.min(y, yMin)
yMax = Math.max(y, yMax)
```

giraffe/src/transforms/band.ts
TCL735: These are backwards because they get set to a different value. For example, the first time is pretty guaranteed to set a new xMin because nothing can be greater than Infinity. Thus xMin gets set and will be set again if the next value is less than the new value.

giraffe/src/transforms/band.ts
TCL735: Yes, I like this. 👍

giraffe/src/transforms/band.ts
TCL735: In this usage, they are the same. I generally avoid ++ unless there is a specific use case that i need.

giraffe/src/transforms/band.ts
```diff
@@ -0,0 +1,210 @@
...
+        }
+        return indices
+      }, [])
+    : []
+
+  return bandLineIndices as number[]
+}
+
+export const getBandName = (fillColumnMap: ColumnGroupMap): string => {
+  return Array.isArray(fillColumnMap.mappings)
```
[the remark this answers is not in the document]
TCL735: 👍

giraffe/src/transforms/band.ts
```diff
@@ -0,0 +1,210 @@
...
+    : []
+
+  bands.forEach(line => {
+    if (minMaxMap[line.lineName][MIN]) {
+      line[MIN] = {...minMaxMap[line.lineName][MIN], fill: line.fill}
+    }
+    if (minMaxMap[line.lineName][MAX]) {
+      line[MAX] = {...minMaxMap[line.lineName][MAX], fill: line.fill}
+    }
+  })
```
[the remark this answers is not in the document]
TCL735: Unfortunately, we can't. The reduce function has to create (and finish) a map first, that will be used in the forEach that follows.

giraffe/src/transforms/band.ts
```diff
@@ -0,0 +1,210 @@
...
+            ...lineData[index],
+            lineName,
+            fill: bandFillColors[bandLines.length],
+          })
+        }
+        return bandLines
+      }, [])
+    : []
+
+  bands.forEach(line => {
```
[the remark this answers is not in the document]
TCL735: 👍

giraffe/src/transforms/band.ts
```diff
@@ -0,0 +1,210 @@
...
+        } else {
+          bandLines.push({
+            ...lineData[index],
+            lineName,
+            fill: bandFillColors[bandLines.length],
+          })
+        }
+        return bandLines
+      }, [])
+    : []
```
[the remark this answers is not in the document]
TCL735: Will do. I have to keep bands as a const below the early return to satisfy TypeScript, which gets annoyed at having bands re-assigned to something that is an array of any in this case.

giraffe/src/types/index.ts
```diff
@@ -286,6 +288,22 @@ export interface LineLayerConfig {
...
+  position?: LinePosition
+  hoverDimension?: LineHoverDimension | 'auto'
+  maxTooltipRows?: number
+  interpolation?: LineInterpolation
+  lineWidth?: number
+  lineOpacity?: number
+  colors?: string[]
+  shadeBelow?: boolean
+  shadeBelowOpacity?: number
+}
```
ebb-tide, line 305:
```suggestion
export interface BandLayerConfig {
  type: 'band' // do not refactor or restrict to LayerTypes.Line
  x: string
  y: string
  fill: string[]
  position?: LinePosition
  hoverDimension?: LineHoverDimension | 'auto'
  maxTooltipRows?: number
  interpolation?: LineInterpolation
  lineWidth?: number
  lineOpacity?: number
  colors?: string[]
  shadeOpacity?: number
  upperColumn: string
  lowerColumn: string
  legendMessage: string
}
```

TCL735: I will be moving "legendMessage" into the top level config since that is where the other legend configurations are located.

TCL735: Also, slight change to "upperColumn" to "upperColumnName" and similarly for "lowerColumn" to "lowerColumnName"

giraffe/src/components/BandLayer.tsx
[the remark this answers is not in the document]
TCL735: Agreed. Less magical and more descriptive.

TCL735: Turns out, hovering for band charts is a LOT more complex. It wasn't just an issue with the 'xy' dimension. All dimensions had issues, including showing incorrect data points.

This is fixed now for all dimensions.

giraffe/src/components/BandLayer.tsx
```diff
@@ -0,0 +1,104 @@
...
+  if (
+    config.hoverDimension === 'auto' &&
+    Object.keys(spec.lineData).length > config.maxTooltipRows
+  ) {
+    hoverDimension = 'xy'
+  } else if (config.hoverDimension === 'auto') {
+    hoverDimension = 'x'
+  } else {
+    hoverDimension = config.hoverDimension
+  }
```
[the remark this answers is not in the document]
TCL735: Yes.

PR #258, opened 2020-08-08
fix: upgrade prismjs in response to security alert

TCL735:
Closes https://github.com/influxdata/giraffe/network/alert/yarn.lock/prismjs/open


TCL735:
Nevermind, this won't work. The offending version is still in the yarn.lock. It is a dependency of a dependency. We will let this one go for now.

PR #262, opened 2020-08-20
fix: improve rendering when lines, mins, and maxes are different lengths

TCL735:
Improvements for #228 

When a line's total number of data points are a different length than the number of mins and maxes for that line, rendering of the band is misaligned. The fix is to fill in empty spots to make them all equal length arrays of data points. This will align the rendering.

Misaligned (BEFORE):
Screen Shot 2020-08-20 at 4 38 25 PM

Fixed (AFTER):
Screen Shot 2020-08-20 at 4 35 40 PM

### DIFF
--- a/giraffe/src/transforms/band.ts
+++ b/giraffe/src/transforms/band.ts
@@ -7,6 +7,7 @@ import {
   BandIndexMap,
 } from '../types'
 import {FILL, RESULT, MAX, MIN} from '../constants/columnKeys'
+import {isDefined} from '../utils/isDefined'
 import {createGroupIDColumn, getNominalColorScale} from './'
 
 export const getBands = (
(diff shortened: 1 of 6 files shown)


giraffe/src/transforms/band.ts
```diff
@@ -126,6 +125,77 @@ export const getBandName = (fillColumnMap: ColumnGroupMap): string => {
   }, '')
 }
 
+export const alignMinMaxWithBand = (
+  lineData: LineData,
+  bandIndexMap: BandIndexMap
+): LineData => {
+  const {rowIndices, maxIndices, minIndices} = bandIndexMap
+
+  for (let i = 0; i < rowIndices.length; i += 1) {
```
drdelambre, line 134: there's a whole lotta variable redeclaration and function creation in these loops here. is the `rowIndicies` number expected to stay small?

giraffe/src/transforms/band.ts
```diff
@@ -126,6 +125,77 @@ export const getBandName = (fillColumnMap: ColumnGroupMap): string => {
...
+    const bandId = rowIndices[i]
+    const minId = minIndices[i]
+    const maxId = maxIndices[i]
+
+    let bandXs = []
```
drdelambre, line 139: decoupling declaration and assignment by hoisting these lets out of the loop will help reduce memory usage on large datasets

hoorayimhelping: +1: in general, we should measure our ui performance before optimizing, but it's a pretty good bet that not creating and destroying memory every iteration of the loop will be more performant than creating it once before the start of the loop.

how big of a loop is this?

PR #263, opened 2020-08-27
fix: allow Raw Data View to parse objects in CSV

TCL735:
Closes idpe #7553

Currently, when an object is returned in a field of the CSV, Raw Data View is unable to parse the CSV correctly especially when the object has commas (multiple properties, nested objects, etc).

The solution is the parse each of the objects correctly, and then re-insert them back into the correct position in the parsed CSV.

NOTE: this is the Giraffe version. Separately, a similar PR will be raised for the `ui` in the near future.


giraffe/src/utils/rawFluxDataTable.ts
```diff
@@ -1,21 +1,98 @@
...
+  CSV_OBJECT_END_STRING,
+} from '../constants/rawFluxData'
 
 export interface ParseFilesResult {
   data: string[][]
   maxColumnCount: number
 }
 
-export const parseFiles = (responses: string[]): ParseFilesResult => {
+export const parseFiles = (
```
drdelambre, line 17: can we make this a new exported function so that we can use the featureflag subsystem in prod to turn it off? creating a "if flag, import NEWparseFiles else import parseFiles" kind of situation? i've just had this problem explode on me too many times.

TCL735: Sure thing, will do.

PR #275, opened 2020-09-10
fix: prevent 24-hour and UTC format from displaying "24" for the hours

TCL735:
Closes idpe #8310

Prevents 24-hour and UTC format without "a" from displaying "24" for the hours

### DIFF
--- a/giraffe/src/utils/formatters.test.ts
+++ b/giraffe/src/utils/formatters.test.ts
@@ -104,7 +104,7 @@ describe('timeFormatter', () => {
     )
   })
 
-  test('24 hour format and UTC format should not show "24" for midnight', () => {
+  test('24 hour format and UTC format without "a" should not show "24" for midnight', () => {
     const utcFormatterWithFormat = timeFormatter({
       timeZone: 'UTC',
       format: 'YYYY-MM-DD HH:mm:ss ZZ',
@@ -118,10 +118,33 @@ describe('timeFormatter', () => {
     let midnight = new Date('2019-01-01T00:00Z')
     expect(utcFormatterWithFormat(midnight)).toEqual('2019-01-01 00:00:00 UTC')
 
+    midnight = new Date('2019-01-01T24:00Z')
+    expect(utcFormatterWithFormat(midnight)).toEqual('2019-01-02 00:00:00 UTC')
+
     midnight = new Date('2019-01-01T23:00Z')
     expect(nonUTCFormatterWithFormat(midnight)).toEqual(
       '2019-01-02 00:00:00 GMT+1'
     )
+  })
+
+  test('24 hour format and UTC format without "a" should display correctly', () => {
+    const utcFormatterWithFormat = timeFormatter({
+      timeZone: 'UTC',
+      format: 'YYYY-MM-DD HH:mm:ss ZZ',
+    })
+    const nonUTCFormatterWithFormat = timeFormatter({
+      timeZone: 'Europe/Berlin',
+      format: 'YYYY-MM-DD HH:mm:ss ZZ',
+      hour12: false,
+    })
+
+    let date = new Date('2019-01-01T15:00Z')
+    expect(utcFormatterWithFormat(date)).toEqual('2019-01-01 15:00:00 UTC')
+    expect(nonUTCFormatterWithFormat(date)).toEqual('2019-01-01 16:00:00 GMT+1')
+
+    date = new Date('2019-01-01T07:00Z')
+    expect(utcFormatterWithFormat(date)).toEqual('2019-01-01 07:00:00 UTC')
+    expect(nonUTCFormatterWithFormat(date)).toEqual('2019-01-01 08:00:00 GMT+1')
   })
 
   test('can format times with format strings', () => {
--- a/giraffe/src/utils/formatters.test.ts
+++ b/giraffe/src/utils/formatters.test.ts
@@ -104,6 +104,26 @@ describe('timeFormatter', () => {
     )
   })
 
+  test('24 hour format and UTC format should not show "24" for midnight', () => {
+    const utcFormatterWithFormat = timeFormatter({
+      timeZone: 'UTC',
+      format: 'YYYY-MM-DD HH:mm:ss ZZ',
+    })
+    const nonUTCFormatterWithFormat = timeFormatter({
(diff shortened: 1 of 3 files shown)


giraffe/src/utils/formatters.ts
```diff
@@ -153,7 +161,11 @@ export const timeFormatter = ({
         }
         return String(Number(lhour))
       }
-      return lhour
+      if (is24hourLocale || UTC_TIME_ZONE.test(timeZone) || hour12 === false) {
+        const numericalHour = Number(lhour) % 24
+        return numericalHour < 10 ? `0${numericalHour}` : `${numericalHour}`
```
Asalem1, line 166: This seems really similar to line 150, but it relies upon a different implementation. I’m curious if we should make them uniform with respect to how they determine if it’s a single digit number or a two digit number

stories/src/csv.stories.tsx
```diff
@@ -39,7 +39,9 @@ storiesOf('CSV from Flux', module)
         'MM/DD/YYYY HH:mm:ss.sss': 'MM/DD/YYYY HH:mm:ss.sss',
         'YYYY/MM/DD HH:mm:ss': 'YYYY/MM/DD HH:mm:ss',
         'YYYY-MM-DD HH:mm:ss ZZ': 'YYYY-MM-DD HH:mm:ss ZZ',
+        'YYYY-MM-DD HH:mm:ss a ZZ': 'YYYY-MM-DD HH:mm:ss a ZZ',
```
Asalem1, line 42: Is it possible to specify ‘a’ with ‘ZZ’?

Asalem1 (approved):
Looks great, just had some thoughts out loud

giraffe/src/utils/formatters.ts
```diff
@@ -153,7 +161,11 @@ export const timeFormatter = ({
         }
         return String(Number(lhour))
       }
-      return lhour
+      if (is24hourLocale || UTC_TIME_ZONE.test(timeZone) || hour12 === false) {
```
hoorayimhelping, line 164: What state of the system do all these conditions being a certain value represent?

```ts
is24hourLocale || UTC_TIME_ZONE.test(timeZone) || hour12 === false
```

in other words, can you store the output of the above expression into a variable that names / describes the situation more eloquently than the three `or` ed expressions?

could you name it something like 

`const isUTCAnd24Hours = is24hourLocale || UTC_TIME_ZONE.test(timeZone) || hour12 === false`

or similar to better clue in what's happening?

giraffe/src/utils/formatters.ts
hoorayimhelping: >I’m curious if we should make them uniform with respect to how they determine if it’s a single digit number or a two digit number

@Asalem1 i think we should not. these are two separate cases and the code _happens_ be similar _now_.  we should let them be separate and see if more instances like this crop up, and if they do, we should make a decision then. two data points that happen to look similar is not enough signal to make a good abstraction.

A bad abstraction is way more deleterious than duplicated code, because duplicated code is just code. An abstraction is a concept in the system that takes mental work to understand and change.

stories/src/csv.stories.tsx
hoorayimhelping: why would you not be able to do this? ZZ says the timezone is UTC

giraffe/src/utils/formatters.test.ts
```diff
@@ -104,6 +104,26 @@ describe('timeFormatter', () => {
...
+    const nonUTCFormatterWithFormat = timeFormatter({
+      timeZone: 'Europe/Berlin',
+      format: 'YYYY-MM-DD HH:mm:ss ZZ',
+      hour12: false,
+    })
+
+    let midnight = new Date('2019-01-01T00:00Z')
+    expect(utcFormatterWithFormat(midnight)).toEqual('2019-01-01 00:00:00 UTC')
+
+    midnight = new Date('2019-01-01T23:00Z')
```
hoorayimhelping, line 121: can we add another test case for midnight being `24`?

can we also add a test case for normal boring cases, for before noon and after noon? e.g. for when the time is like 9 am and when the time is 3 pm?

hoorayimhelping (changes requested):
Couple of nits for clarity

stories/src/csv.stories.tsx
TCL735: This is a storybook change, which is really for testing. I wanted to see what happens if both UTC and am/pm were given in the format by the user. It is for edge case testing. I don't think a serious user would send those options in.

And for your curiosity, am/pm always wins. If 'a' is present, am/pm forces itself into the format regardless of the time zone.

Asalem1: TIL. Thanks

giraffe/src/utils/formatters.ts
TCL735: They are different. In line 150, the format is 'hh' lowercase, which does not display leading zeroes. 'hh' means 1 through 12.

Here, the format is 'HH' uppercase, which can display 0 through 23. We put a leading zero here as a hint to the user that the format goes from 0 to 23.

giraffe/src/utils/formatters.test.ts
TCL735: 1. That is very insightful suggestion. I didn't even realize `Date` can accept "24" in the hours. It can. And it is exactly the kind of test case that addresses this issue. "24" is technically allowed, even though it looks "bad" or "embarrassing" in a time series database.

2. Sure, will do.

stories/src/csv.stories.tsx
hoorayimhelping: >Practically speaking, if a user in the US toggled the UTC timezone, the format of the times shouldn’t have the option of being am/pm

I don't follow why not. UTC and AM/PM are on different axes - they're not mutually exclusive, even though we can't imagine it as a common use case. AM/PM is mutually exclusive with military / 24 hour time, because you can't have 13:15 PM. But as a user in the US, I might want to see my data relative to UTC timezone, but in my American style, so I'd rather see 3 PM than 1500. The timezone of the date shouldn't affect displaying am/pm.

I think the issue here isn't the `ZZ`, it's the `HH`, which renders the time in a 0 padded 24 hour format, which is weird with am / pm

hoorayimhelping (approved):

PR #279, opened 2020-09-15
fix(types): repair Table and Column type to match implementation

sranka:
This PR repairs type definitions to match the implementation
 * Table type changed to match SimpleTable implementation
 * Column type changed to match existing test `parses empty numeric values as null` defined in `fromFlux.test.ts`

These changes were discovered during the optimized integration of influxdb-client-js and giraffe.

### DIFF
--- a/giraffe/src/types/index.ts
+++ b/giraffe/src/types/index.ts
@@ -93,8 +93,8 @@ export interface Formatter {
 
 export interface Table {
   getColumn: GetColumn
-  getColumnName: (columnKey: string) => string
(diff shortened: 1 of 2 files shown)


Asalem1:
I have a quick question about this:

> * Table type changed to match SimpleTable implementation

Is this currently causing us any TS issues on the UI / giraffe?

sranka:
@Asalem, it is causing TS issues when using giraffe outside of InfluxDB UI.

Asalem1 (approved):

PR #282, opened 2020-09-15
feat(util): add queryTable

sranka:
__Proposal:__
This PR adds a new 'queryTable' function that executes flux query (against InfluxDB server) and collects results into a Table instance which can be then directly used in giraffe's Plot. This `queryTable` function is implemented using the official InfluxDB v2 client library, which
- uses fetch [readable stream](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Using_readable_streams) to read data as they go with minimized memory footprint (comparing to receiving the whole CSV response as text)
- allows aborting data retrieval, for example after reaching maximum table rows 
- provides efficient flux response processing, my measurements show up to 50% better processing time than fromFlux(CSV_RESPONSE_TEXT).table
- is fully tested

Moreover, the giraffe rollup distribution completely tree-shakes @influxdata/influxdb-client, since the new code only depends on 'influxdb-client' interfaces.

__Current behavior:__
This library currently does not explain how to get the data to render, it assumes that the user already knows. The fact is that the effective retrieval and processing of response data is not trivial.

__Desired behavior:__
I'd like to have a function that queries InfluxDB and returns a Table that is directly suitable for use in giraffe. This function will also ensure that (by default) at most 100_000 query rows are returned to avoid 'out of memory' issues in the browser.

```
const api = new InfluxDB({url, token}).getQueryApi(org)
const table = await queryTable(api, 'from(bucket: "my-bucket") |> range(start: -30d)' /*, {maxTableRows: 100000} */)
```

__Alternatives considered:__

```
const queryApi = new InfluxDB({url, token}).getQueryApi(org)
const text = await queryApi.queryRaw( 'from(bucket: "my-bucket") |> range(start: -30d)')
const table = fromFlux(text).table
```
works fine, but the solution provided herein is better, because
- it aborts the processing of big results
- it uses an optimized and fully tested code that parses CSV response on the fly

__Use case:__
Why is this important (helps with prioritizing requests)?
From a user perspective, it connects giraffe visualizations to real data from InfluxDB.

__TODO:__
- once we agree that this should be part of giraffe, a top-level readme should be updated with the usage of `queryTable` .

### DIFF
--- a/giraffe/package.json
+++ b/giraffe/package.json
@@ -33,6 +33,7 @@
     "visualization"
   ],
   "devDependencies": {
+    "@influxdata/influxdb-client": "^1.6.0",
     "@rollup/plugin-commonjs": "^13.0.0",
     "@rollup/plugin-node-resolve": "^8.1.0",
     "@types/d3-array": "^2.0.0",
@@ -68,6 +69,7 @@
     "leaflet-ant-path": "^1.3.0",
     "leaflet.markercluster": "^1.4.1",
     "memoize-one": "^5.0.2",
+    "nock": "^13.0.4",
     "node-sass": "^4.14.1",
     "papaparse": "^5.3.0",
     "prettier": "^1.19.1",
--- a/giraffe/package.json
+++ b/giraffe/package.json
@@ -33,7 +33,7 @@
     "visualization"
   ],
   "devDependencies": {
-    "@influxdata/influxdb-client": "^1.6.0",
+    "@influxdata/influxdb-client": "^1.7.0",
     "@rollup/plugin-commonjs": "^13.0.0",
     "@rollup/plugin-node-resolve": "^8.1.0",
     "@types/d3-array": "^2.0.0",
--- a/giraffe/src/index.ts
+++ b/giraffe/src/index.ts
@@ -12,6 +12,7 @@ export {
   timeFormatter,
 } from './utils/formatters'
 export {getDomainDataFromLines} from './utils/lineData'
+export * from './utils/queryTable'
 
 // Transforms
 export {lineTransform} from './transforms/line'
--- a/giraffe/src/utils/newTable.ts
+++ b/giraffe/src/utils/newTable.ts
@@ -12,10 +12,20 @@ class SimpleTable implements Table {
       type: ColumnType
       data: ColumnData
     }
-  } = {}
+  }
 
-  constructor(length: number) {
+  constructor(
+    length: number,
+    columns: {
+      [colKey: string]: {
+        name: string
+        type: ColumnType
+        data: ColumnData
+      }
+    } = {}
+  ) {
     this.length = length
+    this.columns = columns
   }
 
   get columnKeys(): string[] {
@@ -113,7 +123,16 @@ class SimpleTable implements Table {
   }
 }
 
-export const newTable = (length: number): Table => new SimpleTable(length)
+export const newTable = (
+  length: number,
+  columns: {
+    [colKey: string]: {
+      name: string
+      type: ColumnType
+      data: ColumnData
+    }
+  } = {}
+): Table => new SimpleTable(length, columns)
 
 export const newTableFromConfig = (config: Config): Table => {
   if (!config) {
--- a/giraffe/src/utils/newTable.ts
+++ b/giraffe/src/utils/newTable.ts
@@ -22,7 +22,7 @@ class SimpleTable implements Table {
         type: ColumnType
         data: ColumnData
       }
-    } = {}
+    }
   ) {
     this.length = length
     this.columns = columns
@@ -107,9 +107,7 @@ class SimpleTable implements Table {
       )
     }
 
-    const table = new SimpleTable(this.length)
-
-    table.columns = {
+    const table = new SimpleTable(this.length, {
       ...this.columns,
       [columnKey]: {
         name: name || columnKey,
@@ -117,7 +115,7 @@ class SimpleTable implements Table {
         type,
         data,
       },
-    }
+    })
 
     return table
   }
--- a/giraffe/src/utils/queryTable.test.ts
+++ b/giraffe/src/utils/queryTable.test.ts
@@ -1,4 +1,4 @@
-import {queryTable} from './queryTable'
+import {queryTable, DEFAULT_TABLE_OPTIONS} from './queryTable'
 // @influxdata/influxdb-client uses node transport in tests (tests run in node), therefore nock is used to mock HTTP
 import nock from 'nock'
 import {InfluxDB} from '@influxdata/influxdb-client'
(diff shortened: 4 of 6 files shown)


TCL735:
Very intriguing proposal. Will discuss with UI team.

drdelambre:
I really like the use of a generator style interface to progressively parse the csv response! We have some concerns we'd like your viewpoint on:
 1. between this pr and the client library, this effectively creates another csv parser within the ecosystem. as you can imagine in the day to day, this has caused a bunch of problems around termination styles and header formats and the like. bigger picture, having more than one definition of _how a csv turns into a table_ creates a lot of unnecessary overhead as we go chasing the same bugs in 4 places, on 4 timelines, every time we add another this overhead grows exponentially. Is there a path forward where this work can be folded into fromFlux instead?
 1. there's a weird ambiguity about responsibility that this straddles that's difficult to articulate. CSV is a transport mechanism, rooted in decisions i'm not privy to, but one determined more by convention. It's not the only one we can use. JSON is available now, Arrow buffers are on the horizon, basically we aren't married to carrying around csvs as a transport mechanism, but the boundary of this interface proposal is built not on the edges of that transport mechanism, but on a meso layer that's artificially defined in the client library (csv response rows). As we're not dumping csvs into a raw memory buffer that giraffe just uses and are still dealing with a "generate a giant json object" paradigm here, is there a way to better encapsulate the transport mechanism from the parser? This might be the same question as above, might be a change in the client library, just soliciting ideas here.

drdelambre:
I'm feeling like the next step is to scope some affordances into `fromFlux` to take partial csv strings for this to have the data type definitions implemented in the right places. based on this issue https://github.com/influxdata/influxdb-client-js/pull/237 I would say that we should combine (2) and (3), and see if we can get fromFlux to keep some internal state while it's parsing out the object. Then we have condensed the boundaries of the transport mechanism into a single place (giraffe does csv stuff), still leverage the benefits of using a streaming interface, keep backwards compatibility for "fetch it all and process" apps, and keep the datatypes aligned with whatever we're doing in other projects.
thoughts, concerns, pitfalls?

sranka:
A few thoughts. Yes, an approach to minimizing maintenance issues with one tuned implementation is the best when possible. The place where the CSV parsing happens is herein influxdb-client-js, there are low-level functions that help to do this. If these low-level functions shall be combined to provide better usability in the giraffe or other projects, I guess that I should rather proceed in the client library that I am primarily responsible for. I oversee the new `fromFlux` that takes partial CSV strings being semantically `toGiraffeTable(Observable(csv partials)): Promise<Table>` , it belongs to giraffe for sure. Introducing a partial CSV parsing comes with the challenge of changing the existing code from synchronous (possibly asynchronous) to strictly asynchronous style. Backward compatibility with the "fetch it all and process" approach is quite easy to accomplish. Also, the codebase inside influxdata mostly utilizes papaparse for CSV parsing, it might in the end provide a better end-user experience when used with `worker: true` flag, I haven't seen any such usage though.

There is one pitfall of using influxdb-client-js, this library is different when building/bundling for node or browser, this can be (in this case) ignored unless the transport layer of influxdb-client-js is used. It the however the reason why I use 'nock' in the [test](https://github.com/influxdata/giraffe/pull/282/files#diff-e082b8de2c631059b55baa9ab42dd9a1R7) and change the rollup build  of giraffe to target [browser](https://github.com/influxdata/giraffe/pull/282/files#diff-17e9da0990339474f9599d03ad68f1dcR14).

TCL735:
@sranka - I merged a change from this PR separately. https://github.com/influxdata/giraffe/pull/310 

There is a customer who has issues with a build of his project that only shows up in prod env. It might solve his issue.

sranka:
I am willing to help, if there is something that is expected for me to do, pls let know. Should I continue with changing of `fromFlux(string)` internals herein ... or in a new PR?

hoorayimhelping:
I'm coming to this discussion a bit late, but before any more work is invested into this PR, I want to say that I don't think we should do this.

This PR introduces the concept of a network request into Giraffe, something that it has been completely agnostic of. Network requests and data fetching were (and should, by design) be outside the purview of the Giraffe. **Giraffe's responsibility is to render data (specifically time series data) into visualizations.** It should not be opinionated about how the data gets there. It currently only supports flux, as a a consequence of it being a 1st party product that was highly coupled with our system, not as a design decision to only support flux. The library is meant to be agnostic of data fetching, and should only operate on data. This is integration code that belongs at the app level, not the library level.

sranka:
@hoorayimhelping I like the clarity of the giraffe vision that you put into consideration. I also started this PR with one TODO: the stakeholders herein have to agree on whether it should be here. This PR now follows the status quo, in which giraffe also contains parsing of flux results, so it does not change the present situation much. It is still a monolith (1MB transpiled) of various visualization things related to the InfluxDB ecosystem. The new "networking layer" is rather a bridge connecting influxdb-js-client to giraffe. The "networking layer"  (influxdb-client-js) herein does not however contribute to the giraffe distribution (transpiled result), the bridge does. I can also imagine to completely remove the networking layer herein, but the result would be really [hard to use](https://github.com/influxdata/giraffe/pull/282/files#diff-17f5c7a9911986ddd08979a7472cfb3c57ed806b5adebf0e751ffefed88ac926R165), the networking cannot be (in my view) ignored for efficient and mature flux CSV response parsing.

Yes, it might be much better to have a dedicated module/package, where this PR would perfectly fit in. It could be 'giraffe-influx' herein. I can better imagine an introduction of a new package in `influxdb-client-js` (influxdb-client-giraffe) to contain "flux result -> giraffe table" transformation. I am looking for help with such a decision, it will still have a major impact on giraffe users.

hoorayimhelping:
@sranka

Giraffe hasn't really had a strong vision or design (at least since I've been around - it might have before I showed up). We're currently in the process of developing that vision and a roadmap over the next couple of weeks. I appreciate the problem you're trying to solve with this PR, and I think this PR points to a deficiency in giraffe that absolutely exists. The issue is that we're developing other ways we want to solve it that will probably involve deeper changes to giraffe itself.

>This PR now follows the status quo, in which giraffe also contains parsing of flux results, so it does not change the present situation much.

I'm not seeing it that way. Adding an influx library meant to query influx across the network absolutely changes the situation and moves the status quo. If you global search `giraffe/giraffe/src` for an instance of `await` or `then\(`, you won't find any asynchronous processing. This PR adds that. I'm going to draw a line in the sand here and clearly state that asynchronous processing isn't something this rendering library should do.

sranka:
I am OK with the observation that `fromFlux` shouldn't be part of a visualization library and therefore this PR should not land in giraffe. Moving the `flux -> giraffe table` transformation (covered in this PR) to a new `influxdb-client-giraffe` package inside [influxdb-client-js](https://github.com/influxdata/influxdb-client-js/tree/master/packages) is then a solution to an existing design flaw, I have no objections to place it therein ... thoughts/concerns @hoorayimhelping  ? ... or shall I simply wait or postpone this because of plans in the roadmap?

TCL735:
I agree with the recent comments by @hoorayimhelping. Giraffe needs a strong vision and roadmap. And now is not the time to consider this feature. I'm okay with this solution living in another repository.

I will be un-assigning myself from this pull request.

Respectfully, I ask that this pull request be closed once discussion has concluded. Thank you.

sranka:
Thank you for this discussion and your time.

PR #287, opened 2020-09-21
fix(ts): updated the TS definition to include columns

Asalem1:


Asalem1:
There's no point to this PR, gonna close it

PR #297, opened 2020-09-26
fix: use separator for large numbers

### LINKED ISSUE #295: Bug: Gauge value sometimes does not have numeric separators for large numbers
Gauge values follow the localeString format by putting in the appropriate numeric separators for large numbers. For the U.S., the separator is a comma.

Most of the time, the commas are present in large numbers. But sometimes, the comma is missing. This happens when the number includes a decimal portion and the decimal portion ends in zero. Previously, a tangentially related issue - https://github.com/influxdata/giraffe/issues/284 - addressed the position of the needle.

Unresolved is the actual number rendered in the Gauge. It still does not correctly display numbers with decimals that end with 0. It is missing the comma separator.

Here is an example. The rendered number should be **8,264.9530** - with the comma separator
Screen Shot 2020-09-25 at 4 12 15 PM

TCL735:
Closes #295 

Allows large numbers that end with decimal 0 to have numeric separators.
Screen Shot 2020-09-25 at 6 55 58 PM

### DIFF
--- a/giraffe/src/utils/formatStatValue.ts
+++ b/giraffe/src/utils/formatStatValue.ts
@@ -35,19 +35,16 @@ export const formatStatValue = (
   digits = Math.min(digits, MAX_DECIMAL_PLACES)
 
   if (isNumber(value)) {
-    const roundedValueAndDecimal = Number(value)
+    const [wholeNumber, fractionalNumber] = Number(value)
       .toFixed(digits)
       .split('.')
 
-    localeFormattedValue = Number(roundedValueAndDecimal[0]).toLocaleString(
-      undefined,
-      {
-        maximumFractionDigits: MAX_DECIMAL_PLACES,
-      }
-    )
+    localeFormattedValue = Number(wholeNumber).toLocaleString(undefined, {
+      maximumFractionDigits: MAX_DECIMAL_PLACES,
+    })
 
-    if (roundedValueAndDecimal.length > 1) {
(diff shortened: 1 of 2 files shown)


giraffe/src/utils/formatStatValue.ts
```diff
@@ -35,14 +35,20 @@ export const formatStatValue = (
...
   if (isNumber(value)) {
-    const roundedValue = Number(value).toFixed(digits)
-    const endsWithZero = /\.[1-9]{0,}0{1,}$/
-
-    localeFormattedValue = endsWithZero.test(roundedValue)
-      ? roundedValue
-      : Number(roundedValue).toLocaleString(undefined, {
-          maximumFractionDigits: MAX_DECIMAL_PLACES,
-        })
+    const roundedValueAndDecimal = Number(value)
```
hoorayimhelping, line 38: minor suggestion: use [array destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Examples) to make more self documenting code:
```ts
    const [wholeNumber, fractionalNumber] = Number(value).toFixed(digits).split('.')
```

then you can do something like

```ts
if (fractionalNumber) {
  ...
}
```

hoorayimhelping (approved):

giraffe/src/utils/formatStatValue.ts
TCL735: Excellent suggestion. Done.

PR #301, opened 2020-09-29
feat(fromFlux-original-type): added a reference to the original dataType in SimpleTable

Asalem1:
Closes https://github.com/influxdata/ui/issues/123

### Problem

SimpleTable converts the original data type for integers. By losing out on this context, the raw CSV view cannot properly map results based on the CSV string that's returned.

### Solution

Updated the SimpleTable class to include a reference to the originalType in each column. Added a method to the class in order to retrieve the relevant data when needed. After having some brief discussions with @drdelambre I added a default `system` type that is being set as the originalType whenever Giraffe sets a default `number` type since there is no concrete way of knowing whether the actual type is a `long`, `double`, or `unsignedLong`

### DIFF
--- a/giraffe/src/types/index.ts
+++ b/giraffe/src/types/index.ts
@@ -97,10 +97,12 @@ export interface Table {
   getColumn: GetColumn
   getColumnName: (columnKey: string) => string | null // null if the column is not available
   getColumnType: (columnKey: string) => ColumnType | null // null if the column is not available
+  getOriginalColumnType: (columnKey: string) => FluxDataType | null // null if the column is not available
   columnKeys: string[]
   length: number
   addColumn: (
     columnKey: string,
+    originalType: FluxDataType,
     type: ColumnType,
     data: ColumnData,
     name?: string
@@ -129,6 +131,15 @@ export type NumericColumnData =
 export type ColumnData = NumericColumnData | string[] | boolean[]
 
 export type ColumnType = 'number' | 'string' | 'time' | 'boolean'
+
+export type FluxDataType =
+  | 'boolean'
+  | 'unsignedLong'
+  | 'long'
+  | 'double'
+  | 'string'
+  | 'dateTime:RFC3339'
+  | 'system'
 
 export enum LayerTypes {
   RawFluxDataTable = 'flux data table',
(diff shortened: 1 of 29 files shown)


giraffe/src/utils/newTable.ts
```diff
@@ -69,8 +70,19 @@ class SimpleTable implements Table {
...
+    if (!column) {
+      return null
+    }
+
+    return column.originalType
+  }
+
   addColumn(
     columnKey: string,
+    originalType: FluxDataType,
```
drdelambre, line 85: i wonder if we can overload this function to keep backwards compatibility with the existing interface. idk how to do it in typescript, but in plain js it's something like:
```
addColumn(...args) {
   if (args.length === 4) {
         const [ columnKey, type, data, name ] = args
         // try and make something up for the original type
   } else if (args.length === 5) {
         const [ columnKey, originalType, type, data, name ] = args
        // continue like normal
  }
}
```

drdelambre (approved):
changes look good. it'd be nice to see this as a non-breaking semvar change so we can get it out easier. i'll defer to what timmy wants to do here

giraffe/src/types/index.ts
```diff
@@ -97,10 +97,12 @@ export interface Table {
   getColumn: GetColumn
   getColumnName: (columnKey: string) => string | null // null if the column is not available
   getColumnType: (columnKey: string) => ColumnType | null // null if the column is not available
+  getOriginalColumnType: (columnKey: string) => FluxDataType | null // null if the column is not available
   columnKeys: string[]
   length: number
   addColumn: (
     columnKey: string,
+    originalType: FluxDataType,
```
TCL735, line 105: Can we use `fluxDataType` as the property name here? `originalType` does not sound meaningful to me. There is a quick way to replace all instances of that property name in all files using VS Code.

TCL735 (changes requested):

giraffe/src/types/index.ts
Asalem1: Definitely. Great idea

TCL735 (approved):

PR #303, opened 2020-10-06
fix: allow Mosaic to correctly aggregate columns when there are overlapping timestamps

TCL735:
Closes idpe 8443

- allows Mosaic graph to continue aggregating column data when it encounters overlapping timestamps

BEFORE - incorrectly has only 1 column aggregated:
Screen Shot 2020-10-05 at 6 18 08 PM

AFTER - correctly aggregates all columns:
Screen Shot 2020-10-05 at 6 16 03 PM

### DIFF
--- a/giraffe/src/transforms/mosaic.ts
+++ b/giraffe/src/transforms/mosaic.ts
@@ -13,8 +13,8 @@ export const mosaicTransform = (
   fillColKeys: string[],
   colors: string[]
 ): MosaicLayerSpec => {
-  const xInputCol = inputTable.getColumn(xColumnKey, 'number')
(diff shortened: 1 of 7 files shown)


ebb-tide (changes requested):
I think we are still graphing overlapping values. 
We are only `continue`ing when we hit the _first_ retrograde value, rather than skipping all retrograde values. I've refactored the code to be more readable, which brings it out a bit more. I didn't have time to fix it, but I'd be happy to pair tomorrow.

TCL735:
Closing this. Mosaic work turned in new stories because a lot of the Mosaic code is poorly written. Will re-open under a new branch and PR.

PR #309, opened 2020-10-09
chore: update README

TCL735:
- adds defaults for gauge layer configuration
- updates italics style for text regarding defaults


Palakp41 (approved):
👍

PR #315, opened 2020-10-14
fix: make getDomainDataFromLines performant

TCL735:
Closes https://github.com/influxdata/influxdb/issues/18179

- avoid using Array.prototype.concat to repeatedly re-create arrays in a loop
- call `getDomainDataFromLines` only once internally during transformation and save it on the `spec`
- test `getDomainDataFromLines` and `lineTransform` with 5,000,000 data points to ensure error will not be thrown

### DIFF
--- a/giraffe/src/components/LineHoverLayer.tsx
+++ b/giraffe/src/components/LineHoverLayer.tsx
@@ -48,8 +48,7 @@ export const LineHoverLayer: FunctionComponent<Props> = ({
 
   const points = getLineHoverPoints(
     position,
-    spec.lineData,
-    spec.table,
+    spec,
     rowIndices,
     xColKey,
     yColKey,
@@ -111,7 +110,8 @@ export const LineHoverLayer: FunctionComponent<Props> = ({
     fillColKeys,
     fillScale,
     position,
-    spec.lineData
+    spec.lineData,
+    spec.stackedDomainValueColumn
   )
 
   return (
--- a/giraffe/src/utils/lineData.perf.test.ts
+++ b/giraffe/src/utils/lineData.perf.test.ts
@@ -4,7 +4,22 @@ import {lineTransform} from '../transforms/line'
 
 import {largeTable, lineData} from './fixtures/line'
 
+// Baseline: one million data points within 3 seconds
+//   this seems arbitrary and will need adjusting
+const MAX_DURATION = 3000
+
 describe('line graph performance', () => {
+  let start, end
+
+  beforeEach(() => {
+    start = Date.now()
+  })
+
+  afterEach(() => {
+    end = Date.now()
+    expect(end - start).toBeLessThanOrEqual(MAX_DURATION)
+  })
+
   test('getDomainDataFromLines for 1,000,000 data points', () => {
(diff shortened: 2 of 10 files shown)


drdelambre:
how'd you benchmark this?

TCL735:
I attempted to make some sort of performance test for these changes, but ultimately decided not to include the code here. Performance should be measured in a consistent way across the whole of Giraffe rather than piecemeal. As such, we should do performance testing separately.

drdelambre:
oh i wasn't expecting a whole formal testing suite or anything, just curious what criteria you used to judge "this is faster"

TCL735:
For now, fewer calls to `getDomainDataFromLines`. It's called just once rather than 3 or 4 separate times

TCL735:
So, it turns out that adding a test that measures speed is difficult to assert against.

The best we can do for now is to assert that a large data set of 1,000,000 million data points will not throw an error. And we can see roughly how fast it takes. I tried asserting a time limit but could not get it to work reliably, so had to remove the useless assertions.

TCL735:
Okay, it seems to be difficult to assert using Jest's timeout mechanism because `getDomainDataFromLines` is not inherently an async function. It just takes a long time synchronously when the data set is large.

The best we can do here is assert on differences in `Date.now()` from running each test. Seems naive but at least we can establish some sort of arbitrary baseline.

giraffe/src/utils/lineData.perf.test.ts
```diff
@@ -0,0 +1,45 @@
...
+describe('line graph performance', () => {
+  let start, end
+
+  beforeEach(() => {
+    start = Date.now()
+  })
+
+  afterEach(() => {
+    end = Date.now()
+    expect(end - start).toBeLessThanOrEqual(MAX_DURATION)
```
drdelambre, line 20: we should remove this whole test. performance tests that are set up like this have a weird tendency to slowly approach `MAX_DURATION` as the code... matures?.. and then turn flakey as the test runners don't always have the exact amount of resources / quantum dot spins / star alignments as the previous runs, causing some to run slower and some to run faster.

TCL735: Done. Timing expectations removed. We'll settle for just expecting them to handle large data sets without throwing an error.

drdelambre (approved):

giraffe/src/components/LineHoverLayer.tsx
```diff
@@ -48,8 +48,7 @@ export const LineHoverLayer: FunctionComponent<Props> = ({
 
   const points = getLineHoverPoints(
```
Asalem1, line 49: quick question: we have TS to enforce order here, but since we're passing in so many parameters here, wouldn't it be safer to pass in an object that's then deconstructed on the other end? Or is the input too dynamic on the other end?

TCL735: I suggest leaving it for now. This function is only called once. An options object is a good idea in general though.

Asalem1: Makes sense

PR #317, opened 2020-10-15
Improve Legibility of Tooltip

alexpaxton:
### Screenshots

Screen Shot 2020-10-14 at 5 23 45 PM
Screen Shot 2020-10-14 at 4 07 24 PM
Screen Shot 2020-10-14 at 2 20 53 PM

### Changes

- Make background of tooltip darker
- Make font weight of text in tooltip slightly bolder
- Add a handful of color schemes optimized for less series
- Add `legendColorizeRows` option with default `true`
  - If `false` colors are displayed as dots in each row instead of applied to text

### CSS in JS Pattern

- Since we are generating CSS dynamically and the style functions make the component a bit hard to read I moved them to a separate file named `TooltipStyles.ts`
- the `___Styles.ts` naming scheme is probably something we could re-use for other components
- The styles file exports only 1 function which returns an object containing styles for every element organized by element name

### DIFF
--- a/giraffe/src/components/Tooltip.tsx
+++ b/giraffe/src/components/Tooltip.tsx
@@ -2,29 +2,16 @@ import * as React from 'react'
 import {FunctionComponent, useMemo} from 'react'
 import {createPortal} from 'react-dom'
 
-import {TooltipData, Config, ColumnType} from '../types'
+import {TooltipData, Config} from '../types'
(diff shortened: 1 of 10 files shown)


TCL735:
This is great.

When `ui` gets a new version of giraffe with this change, we will have to remember to set a default value for the `legendColorizeRows` config property.

TCL735 (approved):

PR #320, opened 2020-10-16
fix: check for lack of colors in tooltip data

TCL735:
Guard against lack of colors in tooltip data


alexpaxton (approved):
Ahhhh good catch!

PR #330, opened 2020-10-22
feat: allow ticks to be configurable with total count, interval, and starting placement

### LINKED ISSUE #79: Can't control tick generation

TCL735:
Closes #79 

Allows user to specify tick generation in the config.

Screen Shot 2020-10-21 at 8 52 15 PM

### DIFF
--- a/giraffe/README.md
+++ b/giraffe/README.md
@@ -264,7 +264,7 @@ When using the comma separated values (CSV) from the Flux query as the `fluxResp
 
 - **xTickStep**: _number. Optional. **Ignored when xTicks is specified**._ A number representing the tick interval for the x-axis. May be negative.
 
-- **xTickStart**: _number. Optional. **Ignored when xTicks is specified**._ A number representing the value of the first tick on the x-axis. This will determine the placement of all subsequent ticks. Must be within the domain of the x-axis, otherwise will not be rendered. Subsequent ticks may still be rendered if their placement is within the domain.
+- **xTickStart**: _number. Optional. **Ignored when xTicks is specified**._ A number representing a value less than or equal to the first tick on the x-axis. This number will determine the placement of all subsequent ticks. It and any subsequent ticks will be rendered only if they fall within the domain. This number _is_ the value of the first tick when it is in the domain, and at least one of **xTickStep** or **xTotalTicks** is included.
 
 - **yTicks**: _array[number, ...]. Optional._ An array of values representing tick marks on the y-axis. Actual data values and axis scaling may cause Plot to not render all of the given ticks, or Plot rendering may extend beyond all of the rendered ticks. When this option is included, **yTotalTicks**, **yTickStep**, **yTickStart** are ignored.
 
@@ -272,7 +272,7 @@ When using the comma separated values (CSV) from the Flux query as the `fluxResp
 
 - **yTickStep**: _number. Optional. **Ignored when yTicks is specified**._ A number representing the tick interval for the y-axis. May be negative.
 
-- **yTickStart**: _number. Optional. **Ignored when yTicks is specified**._ A number representing the value of the first tick on the y-axis. This will determine the placement of all subsequent ticks. Must be within the domain of the y-axis, otherwise will not be rendered. Subsequent ticks may still be rendered if their placement is within the domain.
+- **yTickStart**: _number. Optional. **Ignored when yTicks is specified**._ A number representing a value less than or equal to the first tick on the y-axis. This number will determine the placement of all subsequent ticks. It and any subsequent ticks will be rendered only if they fall within the domain. This number _is_ the value of the first tick when it is in the domain, and at least one of **yTickStep** or **yTotalTicks** is included.
 
 - **tickFont**: _string. Optional._ The [_CSS font_](https://developer.mozilla.org/en-US/docs/Web/CSS/font) value for the styling of the tick labels and axis labels.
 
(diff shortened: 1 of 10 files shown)


giraffe/src/utils/getTextMetrics.ts
```diff
@@ -18,9 +20,18 @@ export const getTextMetrics = (font: string, text: string): TextMetrics => {
...
 
   span.innerText = text
-
+  spanWithDash.innerText = '-'
+
+  // Text with the same number of characters:
+  //   when it includes a dash (negative number), it tends to be skinnier than
+  //   a positive number because a dash is not as wide as a single digit (most of the time).
+  //   Add padding when text has a dash by making dashes twice as wide.
+  const widthPadding =
```
hoorayimhelping, line 29: is there a cleaner way of doing this? i don't think we can rely on a dash being the same thing as a negative indicator. some platforms might not render `-1` consistently with other platforms

giraffe/src/utils/getTicks.test.ts
```diff
@@ -1,13 +1,293 @@
...
+          valueDomain,
+          VALUE,
+          totalTicks,
+          tickStart,
+          tickStep
+        )
+        expect(result.length).toEqual(0)
+      })
+    })
+  })
```
hoorayimhelping, line 288: great tests from a tactical level. from a higher, strategic level, just reinforcing what i said before - there's a lot of interplay here that is pretty complex and will probably be really confusing for users and might be an indication of larger design issues.

giraffe/src/utils/getTicks.ts
```diff
@@ -112,20 +114,93 @@ export const getTicks = (
...
+  domain: number[],
+  rangeLength: number,
+  tickSize: number,
+  columnKey: string,
+  totalTicks?: number,
+  tickStart?: number,
+  tickStep?: number
+) => {
+  const [start = 0, end = 0] = domain
+  if (isFiniteNumber(totalTicks) || isFiniteNumber(tickStep)) {
```
hoorayimhelping, line 175: should this `||` be an `&&`?

giraffe/src/utils/isFiniteNumber.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    expect(isFiniteNumber(undefined)).toEqual(false)
+    expect(isFiniteNumber(null)).toEqual(false)
+    expect(isFiniteNumber({})).toEqual(false)
+    expect(isFiniteNumber([])).toEqual(false)
+    expect(isFiniteNumber('123')).toEqual(false)
+    expect(isFiniteNumber('-0')).toEqual(false)
+    expect(isFiniteNumber(() => 123)).toEqual(false)
+    expect(isFiniteNumber(Symbol())).toEqual(false)
+  })
+})
```
hoorayimhelping, line 45: for consistency, can you change these `test`s to `it`s please?

hoorayimhelping (approved):
The code looks fine in isolation, but the higher level concerns about the explosion of configuration options still remain for me

giraffe/src/utils/getTextMetrics.ts
TCL735: Yes there is a cleaner way. No extra `<span>` necessary. We do need to check for a dash though, unfortunately. I will send up a new commit in the near future.

giraffe/src/utils/getTicks.ts
TCL735: No. It is correct as `||`. I will demo this to show off the flexibility.

TCL735: Sure, will do.

giraffe/src/utils/isFiniteNumber.test.ts
TCL735: Sure, will do.

giraffe/README.md
```diff
@@ -258,9 +258,21 @@ When using the comma separated values (CSV) from the Flux query as the `fluxResp
...
+
+- **yTotalTicks**: _number. Optional. **Ignored when yTicks is specified**._ A number representing the maximum possible number of ticks to generate on the y-axis. Uses the **yTickStep** as the tick interval if also included. Otherwise the tick interval is taken from dividing the length of the rendered domain by this number. The actual number of rendered ticks may be less than this number due to the size of the tick interval.
+
+- **yTickStep**: _number. Optional. **Ignored when yTicks is specified**._ A number representing the tick interval for the y-axis. May be negative.
+
+- **yTickStart**: _number. Optional. **Ignored when yTicks is specified**._ A number representing the value of the first tick on the y-axis. This will determine the placement of all subsequent ticks. Must be within the domain of the y-axis, otherwise will not be rendered. Subsequent ticks may still be rendered if their placement is within the domain.
```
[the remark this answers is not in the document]
TCL735: Slight clarification and change to the text of `xTickStart` and `yTickStart`.

What do you think of this?
Screen Shot 2020-10-22 at 3 57 30 PM

PR #338, opened 2020-10-28
fix: reconcile generateTicks with the documented behavior in the README

TCL735:
Part of #79 

`generateTicks` should actually generate ticks when the tick start is outside of the domain and the tick step brings any generated ticks within the domain.

### DIFF
--- a/giraffe/src/utils/getTicks.test.ts
+++ b/giraffe/src/utils/getTicks.test.ts
@@ -17,6 +17,23 @@ describe('utils/getTicks', () => {
       const valueDomain = [1, 300]
       const timeDomain = [1603333737559, 1603334037559]
 
+      it('should handle unusual domain values', () => {
+        expect(generateTicks([], VALUE, NaN, NaN, NaN).length).toEqual(0)
+        expect(generateTicks([NaN, NaN], VALUE, NaN, NaN, NaN).length).toEqual(
+          0
+        )
+        expect(
+          generateTicks([null, null], VALUE, NaN, NaN, NaN).length
+        ).toEqual(0)
+
+        expect(generateTicks([100, 0], VALUE, NaN, NaN, NaN).length).toEqual(0)
+
+        expect(generateTicks([100, 99], VALUE, 1, 100, -50).length).toEqual(0)
+
+        expect(generateTicks([100, 100], VALUE, 1, 100, -50).length).toEqual(1)
+        expect(generateTicks([0, 0], VALUE, 1, 100, -0.01).length).toEqual(1)
+      })
+
       it('should generate no ticks when total ticks, tick start, and tick step are not finite numbers', () => {
         expect(generateTicks(valueDomain, VALUE, NaN, NaN, NaN).length).toEqual(
           0
@@ -282,7 +299,7 @@ describe('utils/getTicks', () => {
           tickStart,
           tickStep
         )
-        expect(result.length).toEqual(0)
+        expect(result.length).toEqual(1)
       })
     })
   })
--- a/giraffe/src/utils/getTicks.ts
+++ b/giraffe/src/utils/getTicks.ts
@@ -126,11 +126,16 @@ export const generateTicks = (
   tickStep: number
 ): number[] => {
   const generatedTicks = []
-  const [start, end] = domain
+  const [start = 0, end = 0] = domain
   const stepStart = isFiniteNumber(tickStart) ? tickStart : start
-  const step = isFiniteNumber(tickStep)
-    ? tickStep
-    : (end - stepStart) / (totalTicks + 1)
+
+  let step = tickStep
+  if (!isFiniteNumber(tickStep)) {
+    const parts =
+      isFiniteNumber(totalTicks) && totalTicks !== 0 ? totalTicks : 1
+    step = (end - stepStart) / (isFiniteNumber(tickStart) ? parts : parts + 1)
+  }
+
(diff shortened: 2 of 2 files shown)


giraffe/src/utils/getTicks.ts
```diff
@@ -126,11 +126,16 @@ export const generateTicks = (
...
   const stepStart = isFiniteNumber(tickStart) ? tickStart : start
-  const step = isFiniteNumber(tickStep)
-    ? tickStep
-    : (end - stepStart) / (totalTicks + 1)
+
+  let step = tickStep
+  if (!isFiniteNumber(tickStep)) {
+    const parts =
+      isFiniteNumber(totalTicks) && totalTicks !== 0 ? totalTicks : 1
+    step = (end - stepStart) / (isFiniteNumber(tickStart) ? parts : parts + 1)
```
hoorayimhelping, line 136:
```ts
isFiniteNumber(tickStart) ? parts : parts + 1
```

if `parts` isn't a finite number, what effect does adding 1 to it have? i mean the tests seem to indicate this is fine, but as a random dude reading this, that isn't super clear.

hoorayimhelping (approved):
looks fine, but the denominator to determine `step` is a little confusing to me

Asalem1 (approved):

giraffe/src/utils/getTicks.ts
TCL735: The expression you quoted is checking `tickStart` for being a finite number.

To answer your question, let's briefly go over what tick marks do to a graph. Let's just look at the y-axis as an example. If we have just 1 tick mark on the entire y-axis, this tick mark effectively divides the graph into 2 "parts" - a top half and a bottom half. Having 2 tick marks divides the graph into thirds. 3 tick marks divides into quarters, etc.

So, to answer your question... if `tickStart` is finite, which means the user has specified a number for `tickStart`, we want to divide the graph (ie create a "step") by the number of `parts` which is already determined by this point in the code. If `tickStart` is not finite, which means the user has not specified it, then we want to divide the graph (create a "step") according to algorithm described in the paragraph above. Notice how the "parts" are always 1 more than the number of ticks. That's where the + 1 comes from.

TCL735: This may be difficult to visualize with just words. It's easier to demonstrate by running the code in storybook and using the "User defined ticks" story.

luismedicina23 (commented):
@Main

luismedicina23 (commented):
@MIn

luismedicina23 (commented):
@Main

giraffe/src/utils/getTicks.ts
```diff
@@ -126,11 +126,16 @@ export const generateTicks = (
   tickStep: number
 ): number[] => {
   const generatedTicks = []
-  const [start, end] = domain
+  const [start = 0, end = 0] = domain
   const stepStart = isFiniteNumber(tickStart) ? tickStart : start
-  const step = isFiniteNumber(tickStep)
```
luismedicina23, line 131: `@main
  const step = isFiniteNumber(tickStep)
```

Comment (luismedicina23) [inline giraffe/src/utils/getTicks.ts:131]: @Main
- repeated

**Review by luismedicina23 — COMMENTED** [review]
@main

**Review by luismedicina23 — COMMENTED** [review]
@Mzin

File: giraffe/src/utils/getTicks.test.ts [hunk 3]
```diff
@@ -17,6 +17,23 @@ describe('utils/getTicks', () => {
...
+
+        expect(generateTicks([100, 99], VALUE, 1, 100, -50).length).toEqual(0)
+
+        expect(generateTicks([100, 100], VALUE, 1, 100, -50).length).toEqual(1)
+        expect(generateTicks([0, 0], VALUE, 1, 100, -0.01).length).toEqual(1)
+      })
+
       it('should generate no ticks when total ticks, tick start, and tick step are not finite numbers', () => {
         expect(generateTicks(valueDomain, VALUE, NaN, NaN, NaN).length).toEqual(
           0
```
luismedicina23, line 22: @MIn

giraffe/src/utils/getTicks.test.ts
```diff
@@ -17,6 +17,23 @@ describe('utils/getTicks', () => {
       const valueDomain = [1, 300]
```
luismedicina23, line 17:
```
       No Const valueDomain = [1, 300
```

luismedicina23 (commented):
@TCL735
- Sry, it's my

luismedicina23 (commented):
@Main

giraffe/src/utils/getTicks.test.ts
```diff
@@ -17,6 +17,23 @@ describe('utils/getTicks', () => {
...
+      it('should handle unusual domain values', () => {
+        expect(generateTicks([], VALUE, NaN, NaN, NaN).length).toEqual(0)
+        expect(generateTicks([NaN, NaN], VALUE, NaN, NaN, NaN).length).toEqual(
+          0
+        )
+        expect(
+          generateTicks([null, null], VALUE, NaN, NaN, NaN).length
+        ).toEqual(0)
+
+        expect(generateTicks([100, 0], VALUE, NaN, NaN, NaN).length).toEqual(0)
```
luismedicina23, line 29: giraffe/src/utils/getTicks.test.ts

luismedicina23:
@TCL735 

- Media 6

PR #358, opened 2020-11-04
fix: increase tooltip z-index to be visible with all clockface components

### LINKED ISSUE #357: Tooltip's z-index is too low
Influxdb's UI uses clockface's portal overlay in the Dashboard cell editor. It has a z-index of 9200. Giraffe's tooltip has a z-index of only 400, causing the tooltip to be buried and not visible during cell editing.

Giraffe's tooltip needs a higher z-index as a short term solution. In the long term, z-index should be available as a config option to allow Giraffe consumers to fix this issue without a code change.

TCL735:
Closes #357 

Over 9000!

### DIFF
--- a/giraffe/src/constants/index.ts
+++ b/giraffe/src/constants/index.ts
@@ -130,4 +130,4 @@ export const TOOLTIP_MINIMUM_OPACITY = 0
 
 export const TOOLTIP_MAXIMUM_OPACITY = 1.0
 
-export const CLOCKFACE_Z_INDEX = 10000
+export const CLOCKFACE_Z_INDEX = 9599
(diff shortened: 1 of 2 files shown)


hoorayimhelping (approved):
This is a bit of a blunt instrument type solution. I'd prefer to fix the underlying issue with clockface's z-index but appreciate this is a bugfux. Can you write a follow up ticket in clockface to lower its z-index por favor?

alexpaxton (commented):
I don't think z index can be set higher than 9999

TCL735:
https://github.com/influxdata/clockface/issues/555

TCL735:
> I don't think z index can be set higher than 9999

I tried looking this up and all results say z-index is an int. Seems like it can go over 9999.

But i will go ahead and lower this to 9599, which combined with the existing value will bring it to 9999.

PR #365, opened 2020-11-06
chronograf integration requirements, fix #356

### LINKED ISSUE #356: [GeoWidget] - Leaflet CSS seems to have gone missing
**Build**
Built locally from Master branch. 
Last commit 23a61bb0e5476857ec4dbdaa811eab78398e42c2

**Use Case**
I'm building a test suite for GeoWidget with Giraffe based in NextJS.  
https://github.com/bonitoo-io/giraffe-cypress 

I add GeoWidget elements to simple pages for testing. 

**Expected Behavior**
That the underlying leaflet tiles for open street maps will be composed sensibly. 

**Actual Behavior**
When I open the base case I see the GeoWidget tiles are scattered across the page.  This is most likely caused by missing Leaflet CSS resources. 

**Screenshot**
MissingLeafletCSS01

**What it should look like**
From yesterday's tests
GeoWidget - Basic - Map with Circles -- successfully loads -- after each hook Screenshot

dubsky:
The css is now included in the document on demand.

### DIFF
--- a/giraffe/src/components/Geo.tsx
+++ b/giraffe/src/components/Geo.tsx
@@ -3,7 +3,6 @@ import React, {FunctionComponent, useEffect, useState} from 'react'
 import BingLayer from './geo/bing-maps/Bing'
 import {Map, TileLayer, LayersControl} from 'react-leaflet'
 import Control from 'react-leaflet-control'
-import 'leaflet/dist/leaflet.css'
 
 // Components
 import CircleMarkerLayer from './geo/CircleMarkerLayer'
@@ -64,7 +63,19 @@ const getMinZoom = (width: number): number => {
   return Math.ceil(Math.log2(width / 256) * ZOOM_FRACTION) / ZOOM_FRACTION
 }
 
+const loadLeafletStyles = (loaded => {
+  return () => {
+    if (!loaded) {
+      loaded = true
+      require(['leaflet/dist/leaflet.css'], result => {
+        result.default.use()
+      })
+    }
+  }
+})(false)
+
 const Geo: FunctionComponent<Props> = props => {
+  loadLeafletStyles()
   const {width, height} = props
   if (width === 0 || height === 0) {
     return null
--- a/giraffe/webpack.config.js
+++ b/giraffe/webpack.config.js
@@ -3,10 +3,12 @@ const webpack = require('webpack')
 const pkg = require('./package.json')
 
 module.exports = {
-  entry: './src/index.ts',
+  entry: {
+    index: './src/index.ts',
+  },
   mode: 'production',
   output: {
-    filename: 'index.js',
+    filename: '[name].js',
     libraryTarget: 'umd',
     library: 'Giraffe',
     path: path.resolve(__dirname, 'dist'),
@@ -44,7 +46,15 @@ module.exports = {
       },
       {
         test: /\.s?css$/i,
+        exclude: /leaflet\.css/,
         use: ['css-loader', 'sass-loader'],
+      },
+      {
+        test: /leaflet\.css/,
+        use: [
+          {loader: 'style-loader', options: {injectType: 'lazyStyleTag'}},
+          'css-loader',
+        ],
       },
       {
         test: /\.(png|svg|jpg|gif)$/,
@@ -70,4 +80,9 @@ module.exports = {
       },
     ],
   },
+  plugins: [
+    new webpack.optimize.LimitChunkCountPlugin({
+      maxChunks: 1,
+    }),
+  ],
 }
--- a/giraffe/package.json
+++ b/giraffe/package.json
@@ -86,6 +86,7 @@
     "react-virtualized-auto-sizer": "^1.0.2",
     "s2-geometry": "^1.2.10",
     "sass-loader": "^10.0.3",
+    "style-loader": "^2.0.0",
     "ts-jest": "^24.0.0",
     "typescript": "3.8.3",
     "unraw": "2.0.0",
--- a/giraffe/src/components/GeoLayer.tsx
+++ b/giraffe/src/components/GeoLayer.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {FunctionComponent} from 'react'
+import React, {FunctionComponent, useState} from 'react'
 import {AutoSizer} from 'react-virtualized'
 
 // Components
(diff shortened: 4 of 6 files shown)


giraffe/webpack.config.js
```diff
@@ -70,4 +80,9 @@ module.exports = {
       },
     ],
   },
+  plugins: [
+    new webpack.optimize.LimitChunkCountPlugin({
+      maxChunks: 1,
+    }),
```
hoorayimhelping, line 86: what's up with the chunking optimization? does leaflet do this by default?

giraffe/src/components/Geo.tsx
```diff
@@ -64,7 +63,19 @@ const getMinZoom = (width: number): number => {
...
+const loadLeafletStyles = (loaded => {
+  return () => {
+    if (!loaded) {
+      loaded = true
+      require(['leaflet/dist/leaflet.css'], result => {
+        result.default.use()
+      })
+    }
+  }
+})(false)
```
hoorayimhelping, line 75: why not just load this through a normal import?

hoorayimhelping (changes requested):

giraffe/webpack.config.js
```diff
@@ -3,10 +3,12 @@ const webpack = require('webpack')
 const pkg = require('./package.json')
 
 module.exports = {
-  entry: './src/index.ts',
+  entry: {
```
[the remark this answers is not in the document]
dubsky: This comment applies to all the other questions in this PR.

As a Giraffe user I would be slightly concerned that there is a lot of widgets included but I will use only a few of them on a single page. This dead weight will cause increased load times of the page. This has impact on user experience, google penalizations etc. 

One solution to overcome this issue is to load the code for individual widgets on demand. This would be effective in the case of the geo widget since it is using quite a few external libraries. Right now half the code of giraffe is geo and its libraries and I think that the load time is significant - in my experience it may add even seconds to the load time. 

So I tried to do some investigation to load geo on demand. I was able to build giraffe in a way that there were two separate chunks/entry points but I struggled to incorporate it into an external application in an easy/transparent way. It may even be framework dependant. It didn't have time to resolve it, I plan to return to this once I will be embedding geo into chronograf.

giraffe/webpack.config.js
dubsky: This is a workaround for a webpack problem - once you start using dynamic imports it forces chunking of its output. There might be other conditions when this gets triggered, so I think it is correct to declare that giraffe is just one piece now.

giraffe/src/components/Geo.tsx
dubsky: Optimization that worked for me is that the leaflet css will be embedded into the page only in the case the geo widget is used. It won't be parsed otherwise - it is ~70kb file if I remember correctly. It is better than nothing I think.

giraffe/src/components/Geo.tsx
```diff
@@ -3,7 +3,6 @@ import React, {FunctionComponent, useEffect, useState} from 'react'
 import BingLayer from './geo/bing-maps/Bing'
 import {Map, TileLayer, LayersControl} from 'react-leaflet'
 import Control from 'react-leaflet-control'
-import 'leaflet/dist/leaflet.css'
```
hoorayimhelping, line 6: so if i understand things correctly, importing this file made everything work, and this PR is all the code needed to remove this import?

hoorayimhelping (changes requested):
Does this fix any issues, or is this purely for optimization purposes? @TCL735 and I talked, and it seems like things worked with maps before this change

giraffe/src/components/Geo.tsx
dubsky: Yes - if you have an application which is using giraffe now you have to explicitly import leaflet.css as you mentioned. This is an implementation detail and should be hidden from the user.

TCL735:
There has been a separate PR that updates webpack.config with similar changes. Please see https://github.com/influxdata/giraffe/pull/425 and fix conflicts here.

Do we still need this PR?

karel-rehor:
I've just checked the current build of giraffe with our test suite. 

https://github.com/bonitoo-io/giraffe-cypress

and influxdata/giraffe

```
commit c187384d60d4f4c7b52e5091d8faeeeeab4414fc (HEAD -> master, origin/master, origin/HEAD)
Author: Timmy Luong <TCL735@users.noreply.github.com>
Date:   Wed Dec 16 17:42:15 2020 -0800

    feat: disable Prettier for .md files

```

Leaflet tiles render as they should, but there is an error in the location.  

The tests in the test suites are centered over Futuroscope just north of Poitiers France.

```javascript
const latitude = 46.6671;
const longitude = 0.3700;
```

For some reason the current giraffe code with leaflet.css renders the center of the test map nearer to Tours, which is at a higher latitude.

TracksAndMarkersInfluxdataPR425

I've been writing tests against Dubsky's PR for more than a month.  This is how the data renders after the same test. 

TracksAndMarkersDubskyPR365

For comparison, this is how the coordinates render on Open Street Map's own site. 

OpenStreet046_6671x0_3700

The map should zoom in to Futuropolis. 

OpenStreet046_6671x0_3700_L16

Here's the result of a test run against current Influxdata/giraffe (N.B. this job has been waiting on PR365 to become nightly)

https://ec2-13-56-214-192.us-west-1.compute.amazonaws.com:8443/view/giraffe/job/giraffe-cypress/

Here's the result of yesterday's test run against PR 365

https://ec2-13-56-214-192.us-west-1.compute.amazonaws.com:8443/view/giraffe/job/giraffe-cypress-adhoc/

Note, I've further noticed that without any data presentation layers, current influxdata/giraffe (after PR425) seems to render the correct location.  However, after an SVG or Canvas presentation layer is added, the location seems to be off. PR365 does not have this problem.

dubsky:
> There has been a separate PR that updates webpack.config with similar changes. Please see #425 and fix conflicts here.
> 
> Do we still need this PR?

We need other changes that are not covered by #425 and are part of this PR. I added them later to this one because these I didn't want to create too many PRs for geo-chronograf integration. Perhaps I will create a new PR because this one starts to be difficult to read.

I don't think that the issue reported by @karel-rehor is related by the different approach of #425. I can take a look at it later on.

@hoorayimhelping: Do I understand correctly that you prefer the static import over dynamic/on-demand ?

dubsky:
This pull request is too big/complicated, obsoleted by #425, #437

PR #366, opened 2020-11-06
refactor: remove lodash functions

### LINKED ISSUE #350: Remove usage of lodash functions in giraffe
There are a handful of calls to lodash functions that are unnecessary.

- In `fromFlux` remove: [`get`](https://github.com/influxdata/giraffe/blob/master/giraffe/src/utils/get.ts) and [`groupBy`](https://github.com/influxdata/giraffe/blob/master/giraffe/src/utils/groupBy.ts)
- In the `geo` components remove: [`keys`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys), [`values`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values), and [`indexOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf)

All the above can be replaced with either existing utils or JavaScript built-in functions. The method names are linked to what they can be replaced with.

zoesteinkamp:
closes #350

### DIFF
--- a/giraffe/package.json
+++ b/giraffe/package.json
@@ -92,7 +92,7 @@
     "uuid": "^3.2.1",
     "version-bump-prompt": "^5.0.0",
     "webpack": "^5.2.0",
-    "webpack-cli": "^4.1.0"
+    "webpack-cli": "^4.2.0"
   },
   "peerDependencies": {
     "react": "^16.8.0",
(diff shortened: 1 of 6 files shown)


TCL735:
looks good

TCL735 (approved):

PR #367, opened 2020-11-09
fix: limit tick generation to 1000 ticks

### LINKED ISSUE #362: Tick generation needs a limit
In Giraffe: tick generation when given a small step such as 1 millisecond can cause a bottleneck and freeze up the browser as it tries to create a very large amount of ticks. Even with a reasonable domain, such as 15 minutes, a small step such as 1 millisecond causes the tick generator to attempt to create 1 million ticks or more.

This can occur for any type of axis with a sufficiently large domain. It is easily noticed on a time axis.

There should be a reasonable limit on the tick generator to prevent it from running excessively. Currently, all ticks are labeled. As such, normal usage would likely not exceed 1000 ticks. Probably a lot less in most cases.

TCL735:
Closes #362 

1000 seems like a reasonable limit. This might seem a little bit high but I wanted to leave some wiggle room for any users who might display their graphs on very large screens (a stadium-sized projector) with a small font. Not sure if there would ever be such a use case.

Performance is great with 1000 as the limit. Tested with the existing storybook story on user defined ticks.

 I am open to feedback on this number and willing to adjust accordingly.

### DIFF
--- a/giraffe/src/constants/index.ts
+++ b/giraffe/src/constants/index.ts
@@ -131,3 +131,5 @@ export const TOOLTIP_MINIMUM_OPACITY = 0
 export const TOOLTIP_MAXIMUM_OPACITY = 1.0
 
 export const CLOCKFACE_Z_INDEX = 9599
+
+export const TICK_COUNT_LIMIT = 1000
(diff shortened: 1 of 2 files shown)


hoorayimhelping (approved):
better than standing in a queue

PR #368, opened 2020-11-09
fix: update the size of the sample text that determines axis padding

### LINKED ISSUE #363: Axis padding should account for a thorough set of skinny characters
Axis padding to the left of the y-axis is too small when the sample tick contains a number '1' in the text. This makes the sample tick skinnier than other ticks that have large characters, causing them to become partially hidden.

The sample tick should check against a bigger set of skinny characters to ensure that its width is wide enough to ensure that all ticks' labels will not be hidden.

Here is an example of the problem. Notice the highest tick on the y-axis has a partially hidden label making the '2' difficult to read:
Screen Shot 2020-11-04 at 6 10 55 PM

TCL735:
Closes #363 

Look for more skinny characters when updating the the sample text that determines axis padding. This will improve readability and prevent axis tick labels from being cut off on the y-axis.

### DIFF
--- a/giraffe/src/utils/getTextMetrics.ts
+++ b/giraffe/src/utils/getTextMetrics.ts
@@ -3,6 +3,40 @@ export interface TextMetrics {
   height: number
 }
 
+const addPaddingToSampleText = (text: string): string => {
+  if (typeof text !== 'string') {
+    return text
+  }
+
+  let sample = text
+
+  /* The following are considered skinny characters:
+    -
+    1
+    |
+    i
+    l
+    I
+    !
+  */
+
+  for (let index = 0; index < text.length; index += 1) {
+    const char = text.charAt(index)
+    if (
+      char === '-' ||
+      char === '1' ||
+      char === '|' ||
+      char === 'i' ||
+      char === 'l' ||
+      char === 'I' ||
+      char === '!'
+    ) {
+      sample += char
+    }
+  }
+  return sample
+}
+
 export const getTextMetrics = (font: string, text: string): TextMetrics => {
   const div = document.createElement('div')
   const span = document.createElement('span')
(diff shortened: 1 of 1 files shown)


TCL735:
@hoorayimhelping i think you had mentioned some comments on this previously. adding you to this review to get your take on it.

giraffe/src/utils/getTextMetrics.ts
```diff
@@ -3,6 +3,40 @@ export interface TextMetrics {
...
+    const char = text.charAt(index)
+    if (
+      char === '-' ||
+      char === '1' ||
+      char === '|' ||
+      char === 'i' ||
+      char === 'l' ||
+      char === 'I' ||
+      char === '!'
+    ) {
```
hoorayimhelping, line 33: alternative es6 foxy form:
```ts
const SLENDER_BOIS = ['-', '1', '|', 'i', 'l', 'I', '!']

if (SLENDER_BOIS.includes(char)) {
  sample += char
}
```

hoorayimhelping (approved):
gotta be honest: it's fairly janky, but let's see how it goes

Palakp41 (approved):

giraffe/src/utils/getTextMetrics.ts
TCL735: i like this, will use it.

subirjolly: Quick quesiton. How long is this "text" usually?

TCL735: 10 million characters. No, just kidding. A typical tick label on an axis. For example, "1.00", "1.10". Or even time, "2020-11-09 1:09:33 PM PST". Generally, not very long because it is visual information.

subirjolly: haha. ah i see. thank you!

PR #371, opened 2020-11-09
chore: update README, keywords, and demo page version

TCL735:


hoorayimhelping (approved):
butts status:
- [ ] not held on to
- [x] held on to

Issue #375, opened 2020-11-12
Legend needs to indicate that data set cannot fit inside


TCL735:
Currently, the legend (tooltip) displays information on hover. Usually, all of the data associated with a hover point is displayed. However, there are extreme cases where a hover point can contain hundreds of thousands of points or more. While there is no practical way to view this data, it would be helpful to get some sort of indication that the hover point has more data than can be displayed.

This is related to https://github.com/influxdata/idpe/issues/9107 in a tangential way. That issue involves performance, due to the resources being exceeded. Solving it would still leave the issue of how to notify the user that there is too much information and not everything can be displayed.

PR #376, opened 2020-11-12
fix: limit current hover point's nearest points to the first 200 data points

TCL735:
Closes idpe 9107

When there are many (thousands+) of data points with the same timestamp, hovering over them causes the browser to throw an error - "RangeError: maximum call stack size exceeded". What happens is, Giraffe collates the data of all of the nearest points to the mouse position. And then tries to render all of these points in the legend. When there are too many points, rendering all of these points on update causes this error.

To fix this, we limit the count of the nearest points to the first 200 points. This number was chosen because, visually, the legend won't fit 200 points in typically-used font sizes. Vertically, a laptop screen fits around 60 - 70 rows using a 12px font. So, 200 should be a sufficiently high limit. We can adjust this limit as necessary.

This is a temporary fix, as we should also notify the user, visually, that there are more data points than can be displayed on screen. See #375

### DIFF
--- a/giraffe/src/utils/useHoverPointIndices.ts
+++ b/giraffe/src/utils/useHoverPointIndices.ts
@@ -5,6 +5,8 @@ import {useLazyMemo} from './useLazyMemo'
 import {isDefined} from './isDefined'
 import {minBy} from './extrema'
 
+import {HOVER_POINTS_COUNT_LIMIT} from '../constants'
+
 export const useHoverPointIndices = (
   mode: 'x' | 'y' | 'xy',
   mouseX: number,
@@ -299,7 +301,7 @@ const collectClosestRowIndices = (i0, i1, j0, j1, bins): number[] => {
   let y = j0
 
   while (x <= i1) {
-    closestRowIndices.push(...bins[x][y])
+    bins[x][y].forEach((index: number) => closestRowIndices.push(index))
     x++
   }
 
@@ -307,7 +309,7 @@ const collectClosestRowIndices = (i0, i1, j0, j1, bins): number[] => {
   y++
 
   while (y <= j1) {
-    closestRowIndices.push(...bins[x][y])
+    bins[x][y].forEach((index: number) => closestRowIndices.push(index))
     y++
   }
 
@@ -315,7 +317,7 @@ const collectClosestRowIndices = (i0, i1, j0, j1, bins): number[] => {
   x--
 
   while (x >= i0) {
-    closestRowIndices.push(...bins[x][y])
+    bins[x][y].forEach((index: number) => closestRowIndices.push(index))
     x--
   }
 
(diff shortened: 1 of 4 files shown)


giraffe/src/utils/useHoverPointIndices.ts
```diff
@@ -299,31 +301,31 @@ const collectClosestRowIndices = (i0, i1, j0, j1, bins): number[] => {
...
+    bins[x][y].forEach((index: number) => closestRowIndices.push(index))
     x--
   }
 
   x++
   y--
 
   while (y >= j0) {
-    closestRowIndices.push(...bins[x][y])
+    bins[x][y].forEach((index: number) => closestRowIndices.push(index))
```
hoorayimhelping, line 328: wow, really good eye. this is insidious

giraffe/src/utils/useHoverPointIndices.ts
```diff
@@ -394,8 +396,12 @@ const lookupIndex1D = (
   if (!nearestRows.length) {
     return []
   }
-
-  const nearestDistance = Math.min(...nearestRows.map(d => d.distance))
+  const nearestDistance = nearestRows.reduce((acc, row) => {
+    if (row.distance < acc) {
+      return row.distance
+    }
+    return acc
```
hoorayimhelping, line 403: turn this into a one liner?
```ts
=> Math.min(row.distance, acc)
```

hoorayimhelping (approved):
looks good, but how can we test this or build confidence in this?

giraffe/src/utils/useHoverPointIndices.ts
TCL735: Sure thing.

TCL735:
> looks good, but how can we test this or build confidence in this?

That's a good question. This is problematic in a couple of ways. I attempted to test this using a data set that has around 200,000 rows.

First, I tried to test this manually by putting the data set in a .js file and exporting it as very long string, and then importing this to use as the flux response in Storybook. But doing so causes webpack to take forever to build and run Storybook.

I also tried to test it by using copy/paste from a separate file with the large data set directly into the "custom CSV" field of Storybook. Upon pasting, the browser freezes up as Storybook tries to catch up with the user input in its "custom CSV" knob.

So, it looks like testing through Storybook is not feasible for such an unusual case: very large data set and very unusual data domain (ie, they all have the same time stamp).

I think ultimately, we will just have to see if this error shows up again in the UI with the same error message. Based on the screenshot provided in idpe 9107, the problem was `useHoverPointIndices`. Since we have now limited the output from `useHoverPointIndices` to a scant size, we can see if this resource exceeded issue appears elsewhere in Giraffe code.

TCL735:
As a follow-up, this appears to work with around 67,000 rows. It takes Storybook over 5 minutes to load, but it works. No error is thrown.

hoorayimhelping:
@TCL735 

Cool, thanks for the description of what you did. I'm starting to get the picture that these are all symptoms of the the larger issue with Giraffe. We need to flesh out its API and separate out its constituent parts, I think.

PR #385, opened 2020-11-18
fix: use correct dimensions and hoverable points for Band Plot and untangle unnecessary code

TCL735:
Closes idpe 9144

Band chart was doing strange things with the hover dimensions. In any flux CSV response that comes from using "group by", the hover points become misaligned. To fix this:
- stop doing strange things with the hover dimensions
- define all of the hoverable points for Band chart and use that everywhere
- take advantage of having the line data aligned by band by looking up the corresponding upper and lower sections per band using a matching offset
- fix the aligning of line data by handling edge case where an "upper" or "lower" line is not associated with any "main" line
- memoize the calculation of the hoverable points to make hovering snappy and the user happy

Screen Shot 2020-11-17 at 7 27 54 PM

### DIFF
--- a/giraffe/src/transforms/band.ts
+++ b/giraffe/src/transforms/band.ts
@@ -23,28 +23,29 @@ export const getBands = (
 ): Band[] => {
   const {upperIndices = [], rowIndices = [], lowerIndices = []} = bandIndexMap
   const bands: Band[] = []
-  rowIndices.forEach((rowIndex, position) => {
-    const upperIndex = upperIndices[position]
+  rowIndices.forEach((rowIndex, i) => {
+    if (!isFiniteNumber(rowIndex)) {
+      return
+    }
+    const upperIndex = upperIndices[i]
     const upperIsFinite = isFiniteNumber(upperIndex)
     const upper = upperIsFinite ? lineData[upperIndex] : {fill: ''}
 
-    const lowerIndex = lowerIndices[position]
+    const lowerIndex = lowerIndices[i]
     const lowerIsFinite = isFiniteNumber(lowerIndex)
     const lower = lowerIsFinite ? lineData[lowerIndex] : {fill: ''}
 
-    if (isFiniteNumber(rowIndex)) {
-      const row = lineData[rowIndex]
-      upper.fill = row.fill
-      lower.fill = row.fill
-      const result: Band = {...row}
-      if (upperIsFinite) {
-        result.upper = upper as BandBorder
-      }
-      if (lowerIsFinite) {
-        result.lower = lower as BandBorder
-      }
-      bands.push(result)
+    const row = lineData[rowIndex]
+    upper.fill = row.fill
+    lower.fill = row.fill
+    const result: Band = {...row}
+    if (upperIsFinite) {
+      result.upper = upper as BandBorder
     }
+    if (lowerIsFinite) {
+      result.lower = lower as BandBorder
+    }
+    bands.push(result)
   })
 
   return bands
--- a/giraffe/src/transforms/band.ts
+++ b/giraffe/src/transforms/band.ts
@@ -24,6 +24,9 @@ export const getBands = (
   const {upperIndices = [], rowIndices = [], lowerIndices = []} = bandIndexMap
   const bands: Band[] = []
   rowIndices.forEach((rowIndex, i) => {
+    // Each band must have a "main column" which creates a row index
+    // Any non-main column not associated with a "main" is not a band
+    // Only bands get rendered
     if (!isFiniteNumber(rowIndex)) {
       return
     }
--- a/giraffe/src/transforms/band.ts
+++ b/giraffe/src/transforms/band.ts
@@ -1,76 +1,65 @@
 import {
   BandLayerSpec,
   Band,
+  BandBorder,
   ColumnGroupMap,
   LineData,
   Table,
   BandIndexMap,
 } from '../types'
-import {FILL, LOWER, RESULT, UPPER} from '../constants/columnKeys'
-import {isDefined} from '../utils/isDefined'
+
 import {createGroupIDColumn, getBandColorScale} from './'
 
+import {FILL, LOWER, RESULT, UPPER} from '../constants/columnKeys'
 import {BAND_COLOR_SCALE_CONSTANT} from '../constants'
+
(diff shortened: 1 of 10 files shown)


TCL735:
I added a commit to remove rendering of any "upper" or "lower" columns (lines of data) that are not associated with a "main". Band plot does not allow hovering over orphaned "upper" or "lower" columns, so rendering them would cause confusion.

TCL735:
Ready for review.

TCL735:
Ok, now it's ready for review.

giraffe/src/utils/getBandHoverIndices.ts
```diff
@@ -91,32 +40,58 @@ export const getBandHoverIndices = (
...
 ): BandIndexMap => {
   const bandHoverIndices = {
     rowIndices: [],
     lowerIndices: [],
     upperIndices: [],
   }
+  const hoverGroupBoundaries = {}
+
+  if (Array.isArray(hoverGroupData)) {
+    hoverGroupData.forEach((lineIndex, position) => {
```
hoorayimhelping, line 53: i think i'd like to start standardizing counters to be `i` since it's kind of an industry standard. what do you think about changing `position` to `i`?

giraffe/src/utils/useBandHoverColumns.ts
```diff
@@ -0,0 +1,65 @@
...
+    ys: [],
+    groupColData: [],
+  }
+  const {rowIndices} = bandIndexMap
+
+  rowIndices.forEach(rowIndex => {
+    if (isDefined(rowIndex)) {
+      hoverableColumnData.xs = hoverableColumnData.xs.concat(
+        lineData[rowIndex].xs
+      )
```
hoorayimhelping, line 21: any reason not to go with the more slightly more compact?

```ts
hoverableColumnData.xs = [...hoverableColumnData.xs, ...lineData[rowIndex].xs)
```

giraffe/src/utils/useBandHoverColumns.ts
```diff
@@ -0,0 +1,65 @@
...
+      hoverableColumnData.ys = hoverableColumnData.ys.concat(
+        lineData[rowIndex].ys
+      )
+      groupColData
+        .filter(index => index === rowIndex)
+        .forEach(value => hoverableColumnData.groupColData.push(value))
```
hoorayimhelping, line 27: i think we should prefer `map` and returning the array, rather than pushing to another array in these cases. this is what `map` is specifically designed for.

```ts
hoverableColumnData.groupColData = groupColData
  .filter(index => index === rowIndex)
  .map(value => value)
```

what do you think?

hoorayimhelping (changes requested):

hoorayimhelping:
Shoot, meant to leave a comment: good work on this, a couple of little suggestions, more aimed with larger scaled standardization

giraffe/src/transforms/band.ts
```diff
@@ -1,76 +1,53 @@
...
+  rowIndices.forEach((rowIndex, position) => {
+    const upperIndex = upperIndices[position]
+    const upperIsFinite = isFiniteNumber(upperIndex)
+    const upper = upperIsFinite ? lineData[upperIndex] : {fill: ''}
+
+    const lowerIndex = lowerIndices[position]
+    const lowerIsFinite = isFiniteNumber(lowerIndex)
+    const lower = lowerIsFinite ? lineData[lowerIndex] : {fill: ''}
+
+    if (isFiniteNumber(rowIndex)) {
```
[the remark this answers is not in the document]
TCL735: Good suggestion. I will do the first one. The second one looks really clean, however, it potentially changes the length of the input array, which would throw off the position or `i` when we use it to compare to other arrays which were not similarly filtered.

giraffe/src/utils/getBandHoverIndices.ts
TCL735: Done.

giraffe/src/utils/useBandHoverColumns.ts
TCL735: Yes, normally we would prefer map over forEach. However, in this piece of code after we map, the results have to somehow combine with another array. concat can take care of that for normal arrays. groupColData is a `Float64Array` cannot be an argument to concat.

The other option is to use the spread operator to spread out groupColData's elements into the concat function. But, there could be _many_ elements which causes problems with function calls with too many elements.

So, i think here because of the `Float64Array`, pushing them to an array is as good as it gets.

giraffe/src/utils/useBandHoverColumns.ts
TCL735: I like that. Thanks.

TCL735: Actually, i don't know if this will cause a problem: too many arguments passed to a function call.

This is literal array syntax, so not sure if the underlying implementation is a function call that creates a new array. If so, it will run into the "too many arguments" problem.

TCL735: Looks like it doesn't.

A literal array constructed from de-structuring other arrays uses iterable objects. So, we should be fine going with this suggestion.

giraffe/src/transforms/band.ts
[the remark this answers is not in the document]
hoorayimhelping: @TCL735 fair enough, could you add a comment that position matters here, so we can't filter?

giraffe/src/utils/useBandHoverColumns.ts
hoorayimhelping: @TCL735 can you add a comment explaining those considerations?

hoorayimhelping (approved):
nice work, love the clean up!

couldn't manage to add review comments directly, but basically can you add some comments explaining the special case?

giraffe/src/utils/useBandHoverColumns.ts
TCL735: I ended up doing the destructuring into a literal array, which is backed by iterable objects and should be able to handle as many elements as any array can hold. So, no comment needed here.

giraffe/src/transforms/band.ts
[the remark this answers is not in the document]
TCL735: Sure.

PR #389, opened 2020-11-19
test: ignore 6133 declared but never read errors in stories

hoorayimhelping:
This will let you comment out variable usage without having to comment out declarations when you're writing or testing stories.


TCL735:
So simple yet so amazing.

TCL735 (approved):

Issue #400, opened 2020-11-30
Annotations: on hover a tooltip should appear with title and description


TCL735:
Refer to https://github.com/influxdata/giraffe/issues/377 and https://github.com/influxdata/giraffe/issues/346

On hover, an annotation should display a tooltip with the annotation's title and description.

- Must display a separate tooltip than the tooltip for graph layers
- Tooltip should be anchored to the annotation
- Should work with Giraffe's existing hover dimensions

PR #405, opened 2020-12-01
feat: add AnnotationLayer

### LINKED ISSUE #377: Annotations: Render a single vertical line [optionally: with a note]
### Background
Depends on the ability to render data into layers in Giraffe: See: https://github.com/influxdata/ui/issues/292

It might make sense to do a bit of probing work on 292 (adding Graph Layers) to understand a bit more of how the interface and api of this should work.

#### The product requirements of the UI are:
  - Tooltip should contain timestamp (of that line), the summary text, (if it exists) the message text, and an edit button
  - Annotations inherit their display color from the stream they are a part of

### Acceptance Criteria
- Ability to render a single vertical line the full height of the visualization
  - ~~Hovering over a vertical line displays a tooltip aligned to the vertical line~~ <---- covered by a separate issue
  - The tooltip should be rendered by giraffe, but the content might need to be injectable by the application (in the case of the ui providing an edit button)
- Annotations layers should accept a single color

TCL735:
Closes #377 

**vertical:**
Screen Shot 2020-12-01 at 3 43 36 PM

Screen Shot 2020-12-01 at 3 43 09 PM

**horizontal:**
Screen Shot 2020-12-01 at 3 45 07 PM

Screen Shot 2020-12-01 at 3 45 31 PM


alexpaxton:
Does this include the tooltip bit? Based on the screenshots it is just the line

TCL735:
> Does this include the tooltip bit? Based on the screenshots it is just the line

No, tooltip will be a separate story. I am working on that right now.

giraffe/src/components/AnnotationLayer.tsx
```diff
@@ -0,0 +1,47 @@
...
+  spec: AnnotationLayerSpec
+  config: AnnotationLayerConfig
+}
+
+const ANNOTATION_OVERLAY_DEFAULT_STYLE = {
+  width: '100%',
+  height: '100%',
+}
+
+export const AnnotationLayer: FunctionComponent<AnnotationLayerProps> = props => {
```
hoorayimhelping, line 15: philosophical question: why a `FunctionComponent` over a regular component or a `PureComponent`? i'm not sure if this has changed in the past couple of versions, but my understanding is that a `PureComponent` will only re-render when its props or state changes, whereas a `FunctionComponent` will always re-render and we have to use a fairly complex concept like memoization to make it performant.

giraffe/src/components/AnnotationLayer.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const {height, width, spec, xScale, yScale} = props
+  const annotationsPositions = useMemo(
+    () => getAnnotationsPositions(spec.annotations, xScale, yScale),
+    [spec.annotations, xScale, yScale]
+  )
+
+  return (
+    <svg style={{...ANNOTATION_OVERLAY_DEFAULT_STYLE}}>
```
hoorayimhelping, line 23: is there a way to override this? having `default` in the name implies there's a way to get around the default

giraffe/src/components/AnnotationLayer.tsx
```diff
@@ -0,0 +1,47 @@
...
+            key={i}
+            x1={annotation.startValue}
+            x2={annotation.startValue}
+            y1="0"
+            y2={height}
+            stroke={annotation.color}
+          />
+        )
+      )}
+    </svg>
```
hoorayimhelping, line 45: if i'm understanding this correctly, the annotation spec contains a collection every annotation on the graph? is that right? and this component renders every annotation in the spec?

can there be multiple annotation specs or do all annotations follow the same one? should we separate the concept of an annotation spec from the annotations it specifies (in other words, move the `annotations` collection out of the spec)?

giraffe/src/transforms/annotation.ts
```diff
@@ -0,0 +1,68 @@
...
+): AnnotationLayerSpec => {
+  const [fillColumn] = createGroupIDColumn(inputTable, fillColKeys)
+  const table = inputTable.addColumn(FILL, 'system', 'number', fillColumn)
+  const xCol = table.getColumn(xColumnKey, 'number')
+  const yCol = table.getColumn(yColumnKey, 'number')
+
+  let xMin = Infinity
+  let xMax = -Infinity
+  let yMin = Infinity
+  let yMax = -Infinity
```
hoorayimhelping, line 20: why are the max values negative infinity?

giraffe/src/transforms/annotation.ts
```diff
@@ -0,0 +1,68 @@
...
+      yMax = yCol[i]
+    }
+  }
+
+  const annotations: AnnotationMark[] = []
+
+  if (Array.isArray(inputAnnotations)) {
+    for (let i = 0; i < inputAnnotations.length; i += 1) {
+      if (
+        (inputAnnotations[i].direction === 'y' &&
```
hoorayimhelping, line 45: can you add a comment here explaining what these conditions mean? i get what they're saying, but as a reader, i'm not sure what being within these bounds means.

giraffe/src/types/index.ts
```diff
@@ -494,6 +508,18 @@ export enum SpecTypes {
   Table = 'table',
 }
 
+export interface AnnotationLayerSpec {
+  type: 'annotation'
+  table: Table
+  annotations: AnnotationMark[] // annotations that are renderable (within the domain)
```
hoorayimhelping, line 514: what about calling this something along the lines of `visibleAnnotations` | `renderableAnnotations` since that's what they're representing, right?

hoorayimhelping (changes requested):
Good work. Have some general design questions that relate to higher level giraffe design.

Locally [the annotations over a line graph](http://localhost:50000/?path=/story/annotations--annotations-selectable-marks) story has the annotations fading when I hover. Is that a storybook thing or an annotations thing?

giraffe/src/transforms/annotation.ts
TCL735: Those are initial values. By setting a max with a very small start value (negative Infinity), we guarantee that the first value will become the next max. Same concept for min and using a very large start value.

TCL735:
> Good work. Have some general design questions that relate to higher level giraffe design.
> 
> Locally [the annotations over a line graph](http://localhost:50000/?path=/story/annotations--annotations-selectable-marks) story has the annotations fading when I hover. Is that a storybook thing or an annotations thing?

This should be left for the hovering story to address, as it is hovering behavior that causes this. The hovering story #400 is responsible for making sure hovering works harmoniously with other existing legends.

alexpaxton:
@TCL735 No need for a fade in/out animation on hover, that was just figma eye candy

giraffe/src/components/AnnotationLayer.tsx
TCL735: - The way that AnnotationLayer is used, the props will have changed when it re-renders. The props that change are the mouse coordinates, and they will change if the mouse is hovering over any part of the graph (or AnnotationLayer).

- So, from a re-rendering perspective, it doesn't matter if AnnotationLayer is a function component or a class component. But there is no advantage in being a class component in our use case. Whereas, there are benefits for being a function component. See https://medium.com/@Zwenza/functional-vs-class-components-in-react-231e3fbd7108 under "So why should I use functional components at all?" Additionally, item 4 has a link that leads to a React release blog that mentions the performance impact.

- The memoization is a benefit regardless of whether this is a function component or class component. Even written as a class component, memoization should be done for the positions of the annotation marks.

- And ultimately, i think "re-render" means that the function component is executed, ie the function is called. Actual changes to the DOM is where the performance bottleneck happens. Here, we rely on React to diff and update.

giraffe/src/components/AnnotationLayer.tsx
TCL735: Yes, by de-structuring another set of styles after (to the right of) `ANNOTATION_OVERLAY_DEFAULT_STYLE`, the specific style properties will be overridden.

I did it this way as a clue on how to set configurable styling in the future.

giraffe/src/components/AnnotationLayer.tsx
TCL735: Giraffe's config takes "layers". Our goal is to make these layers work seamlessly together, in other words, we can have multiple layers. Just like having multiple graph layers, we can have multiple annotation layers.

An `AnnotationLayerSpec` or annotation spec is just a blueprint for `PlotEnv` to handle annotation layers.

So, yes, ideally one annotation layer is enough to handle all annotations on the graph. But again, we want Giraffe to be multi-layer. Therefore, annotations can be used as multiple layers.

If that's not the case, then it's a bug.

TCL735:
> @TCL735 No need for a fade in/out animation on hover, that was just figma eye candy

@alexpaxton That wasn't done on purpose. Most likely, the existing hover layers from their respective graph types are causing the annotation line to fade when hovered. In the storybook, the LineLayer's hover layer creates another canvas to draw the hover points and hover line. This additional canvas is most likely the culprit in making the annotation <svg> fade out.

We will fix this as part of #400

giraffe/src/transforms/annotation.ts
TCL735: Sure. I also changed the second condition to explicitly check for 'x' rather than "not y". Subtle distinction but still a distinction, for the sake of being more explicit.

giraffe/src/types/index.ts
TCL735: That's a good point, `annotations` can be confusing because both the config and the spec have that property. Let's use `annotationsData` to be a bit more generic than "visible" or "renderable". The spec is just a place to store valid data.  And this matches up well with LineLayerSpec using "lineData" as a property.

giraffe/src/components/AnnotationLayer.tsx
hoorayimhelping: Without getting into a discussion about individual perferences, that blog post's justifications to use functional components were all based on opinion. In some cases, functional components are easier to test. In other cases, they're much harder to test. In some cases they're less code, in other cases they're more code. And as far as I know, the React team hasn't done anything towards functional component performance updates in the two years since they made those announcements. I think they've backed off the "functions only" train a bit. The point is, it's preference - there are no sound technical reasons to always use one or the other. Sometimes functional components make more sense, sometimes class components and the structure they have make more sense. I bring it up not to say don't use an FC here, but to say 'beware arguments like "the code is cleaner this way"' and also to say that I do not want us to get locked into the mindset that functional components are the only way forward for all our new code.

giraffe/src/components/AnnotationLayer.tsx
hoorayimhelping: >Just like having multiple graph layers, we can have multiple annotation layers.

Is this what we want? We've discussed having annotations just be another layer in the graph, but I don't remember discussing having multiple annotations layers as a separate concept. I want to make sure that the design decisions we make can serve us well in the future

hoorayimhelping (commented):
Let's talk a little more about this this afternoon

giraffe/src/components/AnnotationLayer.tsx
TCL735: What i meant here was that it's flexible enough to take multiple annotations layers. Not necessarily that we encourage the use of multiple annotations layers. If we didn't want multiple layers for annotations, then we simply don't pass in multiple layers. I would not go so far as to explicitly restrict annotations to one layer. That seems like an arbitrary restriction. Unless there is some benefit to doing so that i am not seeing.

giraffe/src/components/AnnotationLayer.tsx
TCL735: Very valid points.

Yes, i do agree that we should look at this from a perspective that is outside of any personal preferences.

My understanding is that class components are almost always more code, because an instance has to be created from a parent class component. In cases where a function component has to rely on a lot of hooks, then maybe the two are closer in terms of "size". But that is probably a case where the component can be broken down further. When a component is reasonably small, I see the functional component as having less code.

But I am certainly aware that there's often a better way to do something. If there is a case where a class component is better at the job, then we'll go with the class component.

I just think that a mindset of defaulting to using class components is not good either. Since that is often a developer's first exposure to React, they tend to fall back on what is comfortable--class components.

hoorayimhelping (approved):

PR #410, opened 2020-12-04
chore: remove fromFluxWithSchema since it is irrelevant

Asalem1:
This PR removes fromFluxWithSchema from the Giraffe library since it is no longer being used by the UI.

As an experimental parser, it was previously only used here:

https://github.com/influxdata/ui/commit/193624242e448a5507352d5dc9058ccd4a766fb2#diff-5d6b0b8e6408adf4b2e8e8dc20deed5aa9a5d7f4599841d610a1777b8694e473L3

and replaced in that same PR. There are no other instances of the function being used in either giraffe or the UI library.


Palakp41 (approved):
Just love the amount of **red** in this pr 👍

PR #415, opened 2020-12-09
Add ability to update lat long based on table data

Palakp41:
Kapture 2020-12-09 at 13 55 09

### DIFF
--- a/giraffe/src/components/Geo.tsx
+++ b/giraffe/src/components/Geo.tsx
@@ -113,6 +113,12 @@ const Geo: FunctionComponent<Props> = props => {
     }
   }
 
+  const latLon = preprocessedTable.getLatLon(0)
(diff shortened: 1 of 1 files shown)


hoorayimhelping (approved):

hoorayimhelping:
@Palakp41 nice work!

If this functionality needs to be published before it can be used, we'll need to coordinate with @TCL735. I think there needs to be some additions to docs before the currently unpublished functionality in giraffe can be properly published

Palakp41:
I don't think we need this published right away. We can wait a few days, we have more work going on in Giraffe that we might need later. Should we write docs about this GEO plot type?

TCL735:
Maps doesn't have any docs.

ivankudibal:
no giraffe component has docs

PR #418, opened 2020-12-10
feat: allow tooltip disable

zoesteinkamp:
We return null for the tooltip if it is disabled

### DIFF
--- a/giraffe/src/components/Tooltip.tsx
+++ b/giraffe/src/components/Tooltip.tsx
@@ -28,6 +28,7 @@ export const Tooltip: FunctionComponent<Props> = ({data, config}) => {
     legendOpacity,
     legendOrientationThreshold: orientationThreshold,
     legendColorizeRows: colorizeRows,
+    legendDisable: disableTooltip,
   } = config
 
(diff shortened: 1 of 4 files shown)


giraffe/src/components/Tooltip.tsx
```diff
@@ -40,9 +41,14 @@ export const Tooltip: FunctionComponent<Props> = ({data, config}) => {
     return TOOLTIP_MAXIMUM_OPACITY
   }, [legendOpacity])
 
-  const columns = columnsWhitelist
-    ? data.filter(column => columnsWhitelist.includes(column.key))
-    : data
+  let columns = []
+  if (disableTooltip != true) {
```
TCL735, line 45: Nitpick: we should use `!==` here. If `disableTooltip` is the value `1` or `"1"` the condition becomes false and the tooltip will be disabled. TypeScript helps here. Being more explicit is good too.

TCL735 (approved):

giraffe/src/components/Tooltip.tsx
zoesteinkamp: Done! Explicit is good!

PR #420, opened 2020-12-10
feat: add hover interaction. add x and y domain values to callback args

### LINKED ISSUE #411: Provide a reverse scaling function that translates xy coordinate positions into time values
### Background
Giraffe provides an `xScale` function to translate timestamps into xy coordinate values that Giraffe can use to render. We want the reverse (inverse). This is useful in Annotations because it translates a click on a graph into a timestamp value that can be saved in our database.

### Acceptance Criteria
- A well tested function to translate xy values into timestamps
- A separate well tested function to translate xy values into y domain values

### LINKED ISSUE #421: Add capability for hover callbacks
### Background
As discussed in a design sync, one way to reduce errors when annotations are created is to give the user direct feedback about the timestamp that they're creating an annotation at. One way of doing this is to provide that information on a tooltip that is always active on hover. To do that, giraffe needs to be able to handle hover callbacks

### Acceptance Criteria
Giraffe can accept a hover callback that is called on hover

hoorayimhelping:
closes #411 
closes #421

### DIFF
--- a/giraffe/src/components/SizedPlot.test.tsx
+++ b/giraffe/src/components/SizedPlot.test.tsx
@@ -73,6 +73,33 @@ describe('the SizedPlot', () => {
         expect(Object.keys(callbackArguments)).toEqual([
           'hoverX',
           'hoverY',
+          'valueX',
+          'valueY',
+          'xDomain',
+          'yDomain',
+          'resetDomains',
+        ])
+      })
+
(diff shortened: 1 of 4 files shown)


TCL735 (approved):

TCL735:
Annotations are chugging along!

PR #424, opened 2020-12-11
fix: no longer will gauge accept thresholds over max or negative

zoesteinkamp:


giraffe/src/components/Gauge.tsx
```diff
@@ -171,8 +171,17 @@ const drawSegmentedGauge = (
   colors: Color[],
   gaugeSize: number
 ): void => {
+  const thresholdColors = []
+  if (Array.isArray(colors)) {
+    colors.forEach(color => {
```
TCL735, line 176: This looks like `filter` rather than `forEach`

giraffe/src/components/Gauge.tsx
```diff
@@ -171,8 +171,17 @@ const drawSegmentedGauge = (
...
+      if (color.value >= minValue && color.value <= maxValue) {
+        thresholdColors.push(color)
+      }
+    })
+  }
+
   const sortedColors = Array.isArray(colors)
-    ? colors.sort((color1, color2) => color1.value - color2.value)
+    ? thresholdColors.sort((color1, color2) => color1.value - color2.value)
```
TCL735, line 184: Since `thresholdColors` is a new variable and guaranteed to be an array, we don't need to check if `colors` is an array anymore. We can just say `sortedColors` is the result of thresholdColors.sort().

TCL735 (changes requested):
See comments.

giraffe/src/components/Gauge.tsx
zoesteinkamp: fixed!

giraffe/src/components/Gauge.tsx
zoesteinkamp: i agree

giraffe/src/components/Gauge.tsx
```diff
@@ -171,9 +171,18 @@ const drawSegmentedGauge = (
...
-    ? colors.sort((color1, color2) => color1.value - color2.value)
-    : []
+  const thresholdColors = []
+  if (Array.isArray(colors)) {
+    colors.filter(color => {
+      if (color.value >= minValue && color.value <= maxValue) {
+        thresholdColors.push(color)
+      }
+    })
+  }
```
hoorayimhelping, line 181: this isn't quite how `filter` is supposed to work.

broadly, there are two types of iteration functions on [Array's prototype.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array) There are functions that iterate over every item of an array like `forEach` and `for... of` and there are functions that iterate over every item of an array and return results, like `map` or `filter`. When to use them comes from intent.

`forEach` is used when you need to do modifications - a good example is looping over every item of an array and converting it into a property of an object.

`map` and `filter` are meant to take an array and make a different array from them. `map` and `filter` both return new arrays.

`filter` works by looping over each item of the array and applying the function that you supply to that item. if the item passes the function's check, it's included in the array. if it fails, it is not included. the function you pass to `filter` should return `true` or `false`, and not have any side effects. calling `push` on an array is a side effect - rather than `push`ing to an outside array, you'll want this function to instead return `true` if the item should be in `thresholdColors` and `false` if it shouldn't . you'll also want to set the value of `threshColors` to be the return value of `filter`:

```ts
let thresholdColors = []
if (Array.isArray(colors)) {
  thresholdColors = colors.filter(// you implement this function to return true or false
```

hoorayimhelping (changes requested):

PR #435, opened 2020-12-17
fix #434 - configs should be read-only

### LINKED ISSUE #434: giraffe should not write into provided input parameters
There is an optimization within PlotEnv.ts which is doing following optimization: On re-render, it takes new config and an old one and is trying to build a new one. This config is reusing objects/parts of the config which haven't changed. 

Unfortunatelly this function is writing into the new configuration. If the configuration is freezed (can be applied by the user or react itself under certain circumastances) the user will get TypeError on the rerender of the plot (identityMerge.ts: line 131).

dubsky:
Fixing defect influxdata#434. Required by geo widget, possibly others as well.

### DIFF
--- a/giraffe/src/utils/identityMerge.ts
+++ b/giraffe/src/utils/identityMerge.ts
@@ -128,7 +128,8 @@ export const setByPath = (target: any, path: Path, value: any): void => {
     currentTarget = nextTarget
   }
 
-  // never write into the input config, Object.freeze might have been applied
+  // beware writing into the currentTarget being part of the input Config passed to
+  // SizedPlot, Object.freeze might have been applied
   const oldValue = currentTarget[path[i + 1]]
   if (oldValue !== value) {
     currentTarget[path[i + 1]] = value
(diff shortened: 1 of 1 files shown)


giraffe/src/utils/identityMerge.ts
```diff
@@ -128,7 +128,11 @@ export const setByPath = (target: any, path: Path, value: any): void => {
     currentTarget = nextTarget
   }
 
-  currentTarget[path[i + 1]] = value
+  // never write into the input config, Object.freeze might have been applied
```
TCL735, line 131: The code is fine, but this comment can be confusing. "never" is not really the case here. I think the code means, if the value hasn't changed, do not write into the config. Because when the value has changed (being not equal), there will be writing into the config.

TCL735 (changes requested):

giraffe/src/utils/identityMerge.ts
dubsky: When the value has changed (being not equal), there will be writing into the config, but not the input one. Rather the 'prevConfig' as referenced in the code if I remember properly.

I have changed the comment to be easier to understand anyway.

TCL735 (approved):

PR #437, opened 2020-12-18
Improvements for geo-chronograf integration

dubsky:
These are changes required before geo widget is merged into chronograf UI.

Replacement for #365


giraffe/src/components/GeoLayer.tsx
```diff
@@ -36,6 +36,7 @@ interface LastRenderProperties {
   zoomOnLastRender?: number
   widthOnLastRender?: number
   heightOnLastRender?: number
+  delayedQueryTimerHandle?
```
Palakp41, line 39: ❓ : Should there be a type on this handle?

giraffe/src/components/GeoLayer.tsx
```diff
@@ -1,5 +1,5 @@
 // Libraries
-import React, {FunctionComponent} from 'react'
+import React, {FunctionComponent, useState} from 'react'
```
Palakp41, line 2: Suggestion: This component is kind of getting too big. Can we break this down into smaller components and move some of this function into a different files, For ex: `utils`?

giraffe/src/components/GeoLayer.tsx
```diff
@@ -92,13 +92,22 @@ const updateQuery = (
...
-    lastRenderProperties.lonOnLastRender = lon
-    lastRenderProperties.zoomOnLastRender = zoom
-    lastRenderProperties.widthOnLastRender = width
-    lastRenderProperties.heightOnLastRender = height
+    if (delayedQueryTimerHandle) {
+      clearInterval(lastRenderProperties.delayedQueryTimerHandle)
+    }
+    // we don't want to call query update too often (on window resize)
+    // we get decoupled from onUpdateQuery call errors
+    lastRenderProperties.delayedQueryTimerHandle = setTimeout(() => {
```
Palakp41, line 100: Suggestion: Can we move this into `utils`? So we can clean this component to render only what it's purpose is.

giraffe/src/components/GeoLayer.tsx
```diff
@@ -122,17 +131,15 @@ const onAutoResize = (
...
-
   updateQuery(
     props,
     lastRenderProperties,
     width,
     height,
-    latOnLastRender === null ? lat : latOnLastRender,
-    lonOnLastRender === null ? lon : lonOnLastRender,
-    zoomOnLastRender === null ? zoom : zoomOnLastRender
+    latOnLastRender === undefined ? lat : latOnLastRender,
```
Palakp41, line 139: Can we clean this up a little bit? I don't think here you would need to have a condition to check for undefined. Instead you should be able to `latOnLastRender || lat` for what you're trying to pass in.

giraffe/src/components/geo/processing/tableProcessing.ts
```diff
@@ -42,45 +42,6 @@ export const preprocessData = (
   return new NativeGeoTable(table, rowLimit)
 }
 
-export const getColumnNames = (
```
Palakp41, line 45: Thanks for cleaning up unused functions! 🎉

Palakp41 (changes requested):
I have added a few suggestions on these components to match our code consistency. Can you take a look at it and let me know if you have any question.

giraffe/src/components/GeoLayer.tsx
dubsky: The problem here is that the typescript runtime target is set to Node.js in giraffe where setTimeout returns NODEJS.Timer, but in browser it returns a number. If I would declare the type correctly, it wouldn't compile. Therefore I left this declaration as is. Seems to be a common problem as I googled.

dubsky: Later I figured I can declare it as ReturnType<typeof setTimeout>.

giraffe/src/components/GeoLayer.tsx
dubsky: The problem here is that lat, lon and zoom are numbers. The cleanup you propose would not work if `latOnLastRender === 0, lonOnLastRender === 0 or zoomOnLastRender ===0` I propose to keep the code as is.

giraffe/src/components/GeoLayer.tsx
dubsky: I don't understant what exactly you would like to move out of this component. Do you mean the setTimeout ? The code you are reffering to is preventing multiple calls to influxdb as the display window is resized. It is an additional condition to prevent querying influxdb with the same parameters repeatedly. I think that my previous comment applies here as well.

giraffe/src/components/GeoLayer.tsx
Palakp41: Here you are checking if `latOnLastRender === undefined`. If it is undefined it returns `lat` but if it will be `latOnLastRender = 0`, the condition for `undefined` is false and it returns `latOnLastRender` which is `0`. How is it different?

karel-rehor:
I've just checked the current state of `dubsky/giraffe` branch `feat/chronograf-geo-integration` against the tests at 

https://github.com/bonitoo-io/giraffe-cypress

dubsky/giraffe latest commit is 

```
commit e07fafd4f4846c8ba7429e6f2dc45d35f092eb73 (HEAD -> feat/chronograf-geo-integration, origin/feat/chronograf-geo-integration)
Author: dubsky <v.dubsky@gmail.com>
Date:   Tue Dec 22 10:48:00 2020 +0100

    chore: improving type definitions

```
The comment that I made in [PR 365](https://github.com/influxdata/giraffe/pull/365#issuecomment-747342826) applies here.  I would open an issue in dubsky/giraffe but do not have permissions. 

Leaflet tiles render as they should, but there is an error in the location.  

The tests in the test suites are centered over Futuroscope just north of Poitiers France.

```javascript
const latitude = 46.6671;
const longitude = 0.3700;
```

For some reason the current giraffe code with leaflet.css renders the center of the test map nearer to Tours, which is at a higher latitude.

TracksAndMarkersInfluxdataPR425

I was writing tests against PR 365 up until the Christmas break.  This is how the data renders after the same test. 

TracksAndMarkersDubskyPR365

For comparison, this is how the coordinates render on Open Street Map's own site. 

OpenStreet046_6671x0_3700

The map should zoom in to Futuropolis. 

OpenStreet046_6671x0_3700_L16

Note, I've further noticed that without any data presentation layers, the map seems to render the correct location.  However, after an SVG or Canvas presentation layer is added, the location seems to be off.

giraffe/src/components/GeoLayer.tsx
dubsky: Let's assume that `latOnLastRender === 0, lat === 100`. The current code returns 0. The shorthand you proposed `latOnLastRender || lat` returns '0 || lat`.  In our example it means `0 || 100` which is evaluated as `100` which is not equal to `0`, the correct result.

giraffe/src/components/GeoLayer.tsx
```diff
@@ -155,11 +162,11 @@ const onAutoResize = (
 }
 
 const GeoLayer: FunctionComponent<OwnProps> = React.memo(props => {
```
Palakp41, line 164: This is the main component `GeoLayer`. All the other things above this line doesn't belong in here. We want to keep our components as specific as possible. The functions above this line can go into a different file.

dubsky: I don't agree with this but it is done. 

The GeoLayer function is based on a proxy pattern. Its only purpose is to stop excessive rendering when it is not needed and all the rendering is then delegated to the `Geo` component. What it does can hardly be anymore specific and I doubt that splitting it into two parts makes it more readable.

PR #438, opened 2020-12-28
build: updating node-notifier

zoesteinkamp:
We are currently getting a warning from dependabot that this library has a security breach, so im updating it to the version with the newest fix

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1251,7 +1251,7 @@
     jest-runtime "^24.9.0"
     jest-util "^24.9.0"
     jest-worker "^24.6.0"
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -1251,7 +1251,7 @@
     jest-runtime "^24.9.0"
     jest-util "^24.9.0"
     jest-worker "^24.6.0"
-    node-notifier "^5.4.2"
+    node-notifier "^8.0.1"
```
hoorayimhelping, line 1254: so this a dependency of `@jest/reporters` version 24.9.0. did you try updating that dependency?

PR #443, opened 2021-01-07
fix: upgrade axios

zoesteinkamp:
We have axios in two places, one is inside storybook-chromatic, which has been depreciated, lets see if upgrading this axios fixes the security issue. If we have to upgrade both that will be a separate PR

### DIFF
--- a/stories/.storybook/config.js
+++ b/stories/.storybook/config.js
@@ -1,6 +1,5 @@
 import {configure, addParameters} from '@storybook/react'
 import {create} from '@storybook/theming'
-import 'chromatic'
 
 addParameters({
   options: {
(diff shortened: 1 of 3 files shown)


TCL735:
I don't think this is the right approach. This might get rid of the alert but the underlying problem is still there. The "unsecure" version of axios is 0.17.1 - which would still be in the yarn.lock.

axios is a dependency of a dependency. The cleanest way to upgrade it would be to upgrade the thing that actually uses axios as the dependency. Ensure that the new version of the thing is also using axios 0.21.1 or higher, because that is the "secure" version that fixes it.

This is probably a storybook upgrade task, which is not straightforward. I have tried it before and storybook doesn't seem to run after upgrading to the latest.

Storybook is not production code, by the way. Just demo code for Giraffe's demo page. "High severity" is an exaggeration for our case.

TCL735 (changes requested):
As discussed, this change won't work because it will be reverted by yarn during a clean install.

TCL735 (approved):

PR #445, opened 2021-01-13
feat: add README explanation for AnnotationLayerConfig and export the interface

### LINKED ISSUE #413: Annotations: add README section for Annotations config
As a developer, I should be able to see written documentation, preferably in the README, for the config options of Annotations.

TCL735:
Closes #413

### DIFF
--- a/giraffe/README.md
+++ b/giraffe/README.md
@@ -289,7 +289,7 @@ When using the comma separated values (CSV) from the Flux query as the `fluxResp
 
 - **height**: _number. Optional._ The height in _CSS px_ of the Plot. Includes the space below the axes surrounding the ticks, tick labels, and axis labels. When not specified, the height of `<Plot>`'s parent element will determine the height.
 
-- **gridColor**: _string. Optional._ The _CSS color value_ of the grid lines. Applies to the inner horizontal and vertical rule lines. Excludes the axes and the border around the graph.
+- **gridColor**: _string. Optional._ The [_CSS color value_](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value) of the grid lines. Applies to the inner horizontal and vertical rule lines. Excludes the axes and the border around the graph.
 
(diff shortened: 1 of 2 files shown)


giraffe/README.md
```diff
@@ -711,7 +711,45 @@ TableGraphLayerConfig uses the `fluxResponse` property from `config` as the data
...
+
+  - **type**: _'annotation'. **Required**._ Specifies that this LayerConfig is an annotation layer.
+
+  - **x**: _string. **Required**._ The column key name of the column that should be visualized on the x-axis. _Annotations must overlay another data set, ie another graph. Therefore, this column key name is from that data set. This should match the overlaid graph's same property._
+
+  - **y**: _string. **Required**._ The column key name of the column that should be visualized on the y-axis. _Annotations must overlay another data set, ie another graph. Therefore, this column key name is from that data set. This should match the overlaid graph's same property._
+
+  - **fill**: _array[string, ...]. **Required**._ An array of column key names of column filters that should be visualized. _Annotations must overlay another data set, ie another graph. Therefore, these column key names are from that data set. This should match the overlaid graph's same property. If the "fill" is not required in the overlaid graph, please explicitly use the overlaid graph's implicit value in AnnotationLayerConfig._
```
jrenee42, line 722: what does 'If the "fill" is not required in the overlaid graph, please explicitly use the overlaid graph's implicit value in AnnotationLayerConfig._' mean?  what would an example of the 'implicit value' be?

TCL735: LineLayer, for example, has **fill** as optional. In the README for LineLayer, fill is described as having a default, or implicit value, if fill is not included. There may other graph types with a similar description for fill.

**fill** is a separate, unrelated problem. It really is required, internally, by Giraffe code. Some graph types allow "fill" to be omitted in the layer config as a way of reducing the number of config properties that the user has to think about.

Unfortunately, AnnotationLayerConfig is not one that should allow omitting "fill" because Annotations must overlay something, which makes Annotations a multi-layer layer. Meaning Annotations cannot be used as a single layer. (Even though, visually, Annotations can be rendered as a single layer.)

hoorayimhelping (approved):

giraffe/README.md
TCL735: @jrenee42 i will put in an example of the "fill" values in a separate PR that addresses default values.

PR #449, opened 2021-01-14
feat: Makes Tooltip disabled when empty array passed to legendColumn

### LINKED ISSUE #386: Ability to disable the legend (tooltip) completely
I am using a graph as a preview inside an overlay and the legend makes the adjacent UI hard to use. Passing in `[]` to `legendColumns` renders an empty tooltip:

image

Proposed property: `disableTooltip: boolean`
Alternatively you could return `null` for the tooltip when `legendColumns` is passed `[]`

**REQUIREMENTS**:

- the tooltip should not render any elements in the DOM if the existing `legendColumns` is an empty array
- the tooltip should not render any elements in the DOM if a new property called `legendDisabled` is `true`
- `legendDisabled: true` takes precedence over `legendColumns`
- `legendDisabled: false` or missing (not included in the config) defers to `legendColumns`

ChitlangeSahas:
Attempts to close #386

### DIFF
--- a/giraffe/src/components/Tooltip.tsx
+++ b/giraffe/src/components/Tooltip.tsx
@@ -43,9 +43,18 @@ export const Tooltip: FunctionComponent<Props> = ({data, config}) => {
 
   let columns = []
   if (disableTooltip !== true) {
-    columns = columnsWhitelist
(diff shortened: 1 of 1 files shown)


zoesteinkamp (approved):
looks good

Issue #447, opened 2021-01-14
Configuration: default values should be consistent and easy to find for all graph types


TCL735:
Currently, configuration default values are all over the place for the different graph types. They are also implemented inconsistently. Some are in an object called **LAYER_DEFAULTS** in the file **giraffe > src > constants > index.ts** while others are implemented by their respective graph type's component. This makes it hard to find and enforce sensible defaults.

Proposal:  all defaults should be easily seen and comparable. This means that the LAYER_DEFAULTS should hold all the default values and no defaults should be implemented by a graph type's component.

PR #459, opened 2021-01-25
refactor: layer default values are consistent and easy to find for all graph types

ChitlangeSahas:
attempts to close: #447 

Following layers are successfully refactored ✅. 

  LineLayer
  HeatmapLayer
  ScatterLayer
  HistogramLayer
  MosaicLayer
  GaugeLayer
  AnnotationLayer
  BandLayer
  TableLayer

### DIFF
--- a/giraffe/src/components/BandLayer.tsx
+++ b/giraffe/src/components/BandLayer.tsx
@@ -27,9 +27,9 @@ export const BandLayer: FunctionComponent<Props> = props => {
   const {config, spec, width, height, xScale, yScale, hoverX, hoverY} = props
 
   const {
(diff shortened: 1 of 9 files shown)


TCL735:
This looks good so far but I think even more can be done, including updating the wording in the README.

Let's leave this PR up for a bit so that more contributions can be made. I have some that i will push up.

TCL735:
This issue is actually a bigger problem and has come up again. The correct way to do this is mentioned in https://github.com/influxdata/giraffe/issues/447  

Closing this PR and will use the above issue as the approach to solve this.

PR #463, opened 2021-01-28
(webpack) giraffe's webpack.config.js updated for scss files

ChitlangeSahas:
The scss files were'nt working properly because of some mis-config in the webpack, this PR attempts too fix it.


TCL735:
Currently, I think a diff will be created when `yarn run build` is run, such as when publishing a new version. Or when running it locally with monitor-ci. The diff could cause confusion.

There are a couple of things we need to do:

- We need to get ahead of it and just commit the current outstanding diff now.

- Also, there should be documentation added somewhere to clarify that the .d.ts files will change any time CSS is changed or added for any component, and `yarn run build` is executed. In any future pull requests, the diffs created from doing so need to be committed.

TCL735 (changes requested):
There are uncommitted diffs for the .d.ts files

ChitlangeSahas:
We had to remove the Prettier Plugin because it doesn't support the `prettierignore` files at the moment. I tested it with the start command and it successfully ignored prettifying the generated files.

stories/package.json
```diff
@@ -36,7 +36,6 @@
     "intl-dateformat": "^0.1.1",
     "memoize-one": "^5.0.2",
     "prettier": "^1.19.1",
-    "prettier-webpack-plugin": "^1.2.0",
```
TCL735, line 39: This should be removed from yarn.lock as well with a `yarn install` and then committed

TCL735 (changes requested):

stories/package.json
ChitlangeSahas: Absolutely, my bad I should have removed it.

TCL735 (approved):

PR #467, opened 2021-02-04
chore: update ts-jest to keep up with recent update to jest

TCL735:
Updates ts-jest to remove this warning:

```ts-jest[versions] (WARN) Version 26.6.3 of jest installed has not been tested with ts-jest. If you're experiencing issues, consider using a supported version (>=24.0.0 <25.0.0). Please do not report issues in ts-jest if you are using unsupported versions.```


hoorayimhelping (approved):
literally 10x

PR #469, opened 2021-02-05
fix: prettierignore file added to the giraffe/giraffe directory.

ChitlangeSahas:
Accidentally missed the `prettierignore` file which was supposed to be a part of a previous PR. Adding that now.


TCL735:
Can you please restore the top level .prettierignore file? Sorry, my bad. We need to keep both.

TCL735 (approved):

PR #481, opened 2021-02-26
fix: don't import the entire damn geo component structure to get a value

hoorayimhelping:
Funny story. I was doing some experimenting with loading Giraffe on the server using a minimal component configuration. And I just couldn't for the life of me understand why I was getting leaflet leaflet errors when I clearly didn't include any leaflet stuff.

I did some digging and it turns out - and this is just _hilarious_ - that some Geo code includes the entire Geo component structure to get a constant that's defined in the `Geo` component and only used in one file. 🙄

### DIFF
--- a/giraffe/src/components/Geo.tsx
+++ b/giraffe/src/components/Geo.tsx
@@ -26,7 +26,6 @@ import {
 import {Config, Table} from '../types'
 
 // Constants
-export const LEAFLET_Z_INDEX = 399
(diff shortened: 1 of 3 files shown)


TCL735:
Glad you found this. I was wondering why unit tests sometimes fail with a similar error message, something about Leaflet, and this is probably the cause of it.

TCL735 (approved):

PR #484, opened 2021-03-03
chore: use single, not double clicks for SizedPlot

jrenee42:
add singleClick to interactionHandlers, and make the SizedPlot respond to single; not doubleclicks

### DIFF
--- a/giraffe/package.json
+++ b/giraffe/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/giraffe",
-  "version": "2.3.0",
+  "version": "2.4.0",
   "main": "dist/index.js",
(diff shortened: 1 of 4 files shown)


hoorayimhelping:
@jrenee42 did you add the updated package.json?

also, semantic nitpick, i think this is probably a `feat`, not a `chore` since we're adding capability

jrenee42:
> @jrenee42 did you add the updated package.json?
> 
> also, semantic nitpick, i think this is probably a `feat`, not a `chore` since we're adding capability

good catch.  just added package.json; and changed chore: to feat: for the first commit too.

hoorayimhelping (approved):
Awesome!

PR #487, opened 2021-03-04
chore: export the TimeFormatterFactoryOptions interface

TCL735:
As a follow-up to https://github.com/influxdata/ui/issues/768 and to allow consuming apps to improve time formatting, expose the options that are configurable.

### DIFF
--- a/giraffe/package.json
+++ b/giraffe/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/giraffe",
-  "version": "2.4.0",
+  "version": "2.5.0",
   "main": "dist/index.js",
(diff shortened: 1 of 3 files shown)


hoorayimhelping (approved):
LGTM, but I think we should start bumping the package version when we make changes. it leads to a slightly more self documenting release because the commit hash of the release points to the entire change (like this: https://github.com/influxdata/giraffe/releases/tag/2.4.0) rather than to the package bump (like this: https://github.com/influxdata/giraffe/releases/tag/v2.1.0)

TCL735:
Updated with the version bump in package.json

PR #505, opened 2021-03-20
fix: raw table +Inf and -Inf renders as such

ChitlangeSahas:
Closes: https://github.com/influxdata/ui/issues/804

### DIFF
--- a/giraffe/src/utils/fromFlux.ts
+++ b/giraffe/src/utils/fromFlux.ts
@@ -277,6 +277,13 @@ const parseValue = (value: string | undefined, columnType: ColumnType): any => {
   }
 
   if (columnType === 'number') {
+    if (value === '+Inf') {
+      return '+Inf'
+    }
+
+    if (value === '-Inf') {
+      return '-Inf'
+    }
     return Number(value)
   }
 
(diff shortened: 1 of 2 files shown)


TCL735:
Strangely, this issue is not present with Giraffe's own implementation of Table Graphs. I'm looking at why. There might be a tangential issue for Giraffe regarding how it parses tables. At first glance, it appears to parse them in two different ways but only uses one of those ways.

I will continue looking at the parsing issue for a bit before I approve this.

giraffe/src/utils/fromFlux.ts
```diff
@@ -277,6 +277,13 @@ const parseValue = (value: string | undefined, columnType: ColumnType): any => {
   }
 
   if (columnType === 'number') {
+    if (value === '+Inf') {
+      return '+Inf'
+    }
+
+    if (value === '-Inf') {
+      return '-Inf'
+    }
```
TCL735, line 286: This solution won't last forever. What if there is another value added in the future that can't be forced into a number? I think, ideally, the number column should just gracefully treat non-numbers as the original string value. Something like this:

```
  if (columnType === 'number') {
    const parsedValue = Number(value)
    return parsedValue === parsedValue ? parsedValue : value
  }
```

In the `return` statement, NaN will fail the first condition and the original string will be returned, which will be whatever was in the flux CSV

TCL735 (changes requested):

PR #510, opened 2021-03-25
feat: implement foundational component for Static Legend

### LINKED ISSUE #492: Graph Legend: Position Legend next to Graph

TCL735:
Closes #492 

Initial work on the Static Legend


TCL735:
Don't merge this yet. There is a maxium call stack size exceeded error when the graph gets hovered. Looking into it now.

TCL735:
Ok, the issue has been fixed. Ready for review.

TCL735:
A quick commit to add a className to the static legend.

giraffe/src/components/Plot.tsx
```diff
@@ -51,10 +72,27 @@ export const Plot: FunctionComponent<Props> = ({
             </SizedTable>
           )
         }
+        if (config.staticLegend) {
+          resized = resizePlotWithStaticLegend(
```
jrenee42, line 76: create a method that takes the resized variable to create the two different sizedPlotWith Legend in one place; DRY

stories/src/staticLegend.stories.tsx
```diff
@@ -0,0 +1,249 @@
...
+  tickFontKnob,
+  showAxesKnob,
+  interpolationKnob,
+  timeZoneKnob,
+  tooltipOrientationThresholdKnob,
+  tooltipColorizeRowsKnob,
+} from './helpers'
+
+const maxValue = Math.random() * Math.floor(200)
+
```
jrenee42, line 27: add documentation; of a minimalist example; ie:

 the Plot needs a 'staticLegend' config object to show the static legend;  this is the minimum needed:

```
 const staticLegend =  {
    heightRatio: .2,
        border: '1px solid orange',
        fontBrightColor: 'yellow',
  }
```

to be added to the plot's config object

giraffe/README.md
```diff
@@ -374,27 +374,61 @@ When using the comma separated values (CSV) from the Flux query as the `fluxResp
...
+
+- **legendOpacity**: _number. Optional. Defaults to 1.0 when excluded._ The [_CSS opacity_](https://developer.mozilla.org/en-US/docs/Web/CSS/opacity) of the legend and tooltip. 0 means the legend and tooltip are invisible, while 1.0 means the legend and tooltip are solid and their content may cover anything underneath. May be overriden.
+
+- **legendOrientationThreshold**: _number. Optional. Defaults to undefined when excluded._ The number of columns in the legend that will determine the direction of columns in the legend. When _undefined_ or when the total number of columns is less than or equal to it, the columns in the tooltip will display horizontally. When the total number of columns is greater, the columns will display vertically. May be overriden.
+
+- **staticLegend**: _Object. Optional._ An object with properties that override the **legend** properties and apply only to the static legend that is always visible. Has a few properties that are unique to the static legend. Properties are:
```
jrenee42, line 403: for clarity:  amend to: 
Include this object to always show a static legend, which is an object with properties that override the **legend** properties and apply only to this static legend . Has a few properties that are unique to the static legend; which includes heightRatio; the most important of the properties.

giraffe/README.md
```diff
@@ -374,27 +374,61 @@ When using the comma separated values (CSV) from the Flux query as the `fluxResp
...
+
+  - **font**: _string. Optional. Defaults to '10px monospace' when excluded._ The [_CSS font_](https://developer.mozilla.org/en-US/docs/Web/CSS/font) value for the styling of the static legend.
+
+  - **fontBrightColor**: _string. Optional. Defaults to #f6f6f8 when excluded._ The [_CSS color value_](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value) of any text that is not a column header or in a row inside the static legend.
+
+  - **fontColor**: _string. Optional. Defaults to #bec2cc when excluded._ The [_CSS color value_](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value) of the column headers in the static legend. The rest of the static legend will use the color scheme set by the `LayerConfig`'s `colors` options.
+
+  - **heightRatio**: _number. Optional. Defaults to 0 when not included._ A fraction expressed as a decimal to indicate the height of the static legend which is a retangular box directly beneath the graph and axes. Must be between 0 and 1 inclusive. Values outside of this range are considered 0.
```
jrenee42, line 423: this should either:
default to .2 ;
or be required.

please add:  must set this to a non-zero number to actually show the static legend.

jrenee42 (changes requested):
found one instance to remove duplicated code, and requested documentation changes
(plus making the heightRatio either required or have a default of .2

the default of 0 is confusing)

giraffe/src/components/Plot.tsx
TCL735: The width and heights are different though. I think DRY does not need to be applied fanatically even in cases where similar code is only repeated once. DRY is more for 3+ uses of something. We don't reach 3 here.

giraffe/README.md
TCL735: Good point about needing this object in order to show a static legend. I will update the wording.

giraffe/README.md
TCL735: We'll default to 0.2 for now. Agree that 0 doesn't make any sense because we want users to have less configuration to worry about.

stories/src/staticLegend.stories.tsx
TCL735: I applied to 0.2 ratio as a default rather than comment it in the storybook, which doesn't always interest everyone.

Agree that 0 as a default height makes no sense. Once a user has a `staticLegend` property, they should get a reasonable height as the default.

giraffe/src/components/Plot.tsx
jrenee42: i disagree, at the very least the two instances of it were confusing to me.

stories/src/staticLegend.stories.tsx
jrenee42: hmm...if there are defaults for everything, if the user accepts all the defaults, then what does the staticLegend object look like?

an empty object?

or if they accept all defaults;
then something like:

```
const staticLegend = {show: true}
```
so something is in there.......
thoughts?

giraffe/src/components/Plot.tsx
TCL735: The piece of code here is a simple `if` without an `else`.  And inside the `if` block is 2 statements.

^ That gets repeated once. It's two total usages. Each usage has different widths and heights. I don't think this is very confusing.

TCL735: My point is: DRY should not be fanatically applied. There should be some leeway with regards to not making everything into a function.

If the above piece of code were more complex, I would agree that DRY makes sense. Here, it's just unnecessary and just adds function bloat.

jrenee42 (approved):

Issue #516, opened 2021-03-26
Annotations: Tweak Click Target


hoorayimhelping:
The click target is kind of hard to click on. Let's tweak it to make it easier.

ChitlangeSahas:
Current finding: 

The little div element is getting in the way of our click target. 

![image](https://user-images.githubusercontent.com/18511823/112695607-a78d3600-8e41-11eb-9cad-9ef52534dbc0.png)

There are a couple of solutions possible, one is to move the tooltip up by 5 more pixels, but that seems hacky so I will try shortening the div that is covering it since it looks like half of it is empty anyway. I'll also try to mess with the Z-index and see if that helps.

PR #517, opened 2021-03-26
fix: make annotation target better clickable

ChitlangeSahas:
closes: #516 

As described in the comments in the above issue, the `div` was in way of the click target, so it would not be clickable in that clipped area. The proposed solution is to decrease the `tooltipCaret` size, and render the `target` above the `Annotation-Line`

`div` no longer in the way: 

Screen Shot 2021-03-26 at 3 09 10 PM

what it looks like in UI:

When hovering on the line: <pointer shows, line dims color, indicating it's clickable> 

Screen Shot 2021-03-26 at 5 57 54 PM


daviddkkim:
@ChitlangeSahas could we get a screenshot of what this looks like?

daviddkkim:
@ChitlangeSahas  

awesome- yeah, I think the smaller caret size is fine.

Would it be too much of a scope creep to add the same onClick behavior of the "clicktarget" to the actual annotation line as well? So that the click target and the annotation line are both clickable.

ChitlangeSahas:
> @ChitlangeSahas
> 
> awesome- yeah, I think the smaller caret size is fine.
> 
> Would it be too much of a scope creep to add the same onClick behavior of the "clicktarget" to the actual annotation line as well? So that the click target and the annotation line are both clickable.

I can do that!!

ChitlangeSahas:
> @ChitlangeSahas
> 
> awesome- yeah, I think the smaller caret size is fine.
> 
> Would it be too much of a scope creep to add the same onClick behavior of the "clicktarget" to the actual annotation line as well? So that the click target and the annotation line are both clickable.

I can give it a shot 👍, I don't believe it would be too hard. What do you think about the annotation line getting thicker just a tad when we hover on it, suggesting that its clickable. Or perhaps change shade?

Barbara mentioned it being not easy to recognize that it was clickable?

daviddkkim:
Instead of thicker (since there were concerns of it being hard to click/distinguish when there's multiple annotations crowded), we can try the classic hover interaction of the hovered item's color getting lighter

daviddkkim:
c-twilight: #fad9ff on hover

daviddkkim:
Alright - the color change on hover is a scope creep to this issue since this one is about making the target clickable. 
let's not worry about that in this issue  - don't let me block you w a scope creep!!

ChitlangeSahas:
> Alright - the color change on hover is a scope creep to this issue since this one is about making the target clickable.
> let's not worry about that in this issue - don't let me block you w a scope creep!!

I figured out the hover-color-change, if it's desirable and Bucky and you like it we can keep it else I can delete it also.

hoorayimhelping (approved):
awesome work!

daviddkkim:
@ChitlangeSahas and I chatted about the hover state of the line and the polygon not being shared. It's good as is from my pov - we can figure out the issue later!

daviddkkim (approved):
lgtm

PR #526, opened 2021-04-06
fix: update the doc link

### LINKED ISSUE #519: Geo: Fix link when results are truncated
### Background

Investigate, what is determining that results are truncated? Is it flux or something internally?
The code where you can find this is: 
https://github.com/influxdata/giraffe/blob/master/giraffe/src/components/Geo.tsx#L227

Currently the Geo component sends the wrong link when results are truncated.

See https://github.com/influxdata/ui/issues/997

###Acceptance Criteria
- The link should be updated to point to: https://docs.influxdata.com/influxdb/cloud/visualize-data/visualization-types/map/

zoesteinkamp:
closes #519 
Alright so with this link updated this issue is closed. Also i checked with the truncated limits and although mapbox does not have any specific defined limits in the docs. And i did test the marker map since thats currently active map and it seems like beyond 2000 led to lag. I am writing and linking a separate issue to determine if we cause the lag or if its mapbox

### DIFF
--- a/giraffe/src/components/Geo.tsx
+++ b/giraffe/src/components/Geo.tsx
@@ -226,7 +226,7 @@ const Geo: FunctionComponent<Props> = props => {
           >
             Results are truncated.
             <a
-              href="https://docs.influxdata.com/chronograf/latest/guides/geo-widget#data-downsampling"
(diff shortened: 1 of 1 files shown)


Palakp41 (approved):
Looks good!

Issue #525, opened 2021-04-06
Annotation: Clamp maximum width of tooltip to somewhere between 200-300 px


hoorayimhelping:
### Background
Tooltips with a lot of text in them can extend out very far. We want to clamp the maximum width to somewhere between 200-300 px (we're not sure the exact number, it'll take some tweaking)

https://github.com/influxdata/ui/issues/1121

<img src="https://images.zenhubusercontent.com/5c33b08d24dda4712ef4ae10/f4f7b921-0310-408d-bd87-ad8e92c048d9" width="80%" height="80%" />

### Acceptance Criteria
- Tooltips are clamped to a maximum width
- The carets are properly centered within this width

ChitlangeSahas:
I put up a PR for the toolTip caret positioning, I still agree that we should limit the tooltip box width.

PR #528, opened 2021-04-07
fix: use better positioning for the AnnotationTooltip caret and add maxWitdh to tooltip

ChitlangeSahas:
closes: #525
addresses: https://github.com/influxdata/ui/issues/1121

we realized that we need a way to handle the tooltip caret position, [issues:  #525, https://github.com/influxdata/ui/issues/1121]. This PR attempts to provide a solution. Following is a brief about of how how we ended up with this solution. 

First we thought that the issue in https://github.com/influxdata/ui/issues/1121 would be solved by limiting the width of the tooltip box. But we quickly realized that there are some edge cases (literally) which we don't think would be covered by just limiting the size of the tooltip.

For example, this annotation in the corner renders its caret, in between the two lines, because at it's current state its programmed to render the caret in the middle of the tooltip box.

image

To address this, one solution is to re-think the caret position. Currently, we position the caret in the center of the tooltip box, if we position it on top of the annotation line instead, we can handle those edge-case corner renders. This is what it looks like after the change:

Screen Shot 2021-04-06 at 10 49 03 PM

Screen Shot 2021-04-06 at 10 48 57 PM

In the UI with a long name:

Screen Shot 2021-04-08 at 10 54 57 AM

### DIFF
--- a/giraffe/src/components/AnnotationTooltip.tsx
+++ b/giraffe/src/components/AnnotationTooltip.tsx
@@ -50,7 +50,7 @@ export const AnnotationTooltip: FunctionComponent<Props> = props => {
   )
 
   let tooltipCaretStyle: CSSProperties = {
-    position: 'absolute',
+    position: 'fixed',
     borderWidth: '7px',
     borderStyle: 'solid',
     borderColor: 'transparent',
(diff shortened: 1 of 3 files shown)


daviddkkim (approved):
@ChitlangeSahas 
you beautiful human being!

hoorayimhelping (approved):
Nice work dude! This is great.

We'll release once we get another commit that addresses 1121

TCL735:
The top border of the tooltip looks cut off in Giraffe's storybook. May I suggest adjusting the margins in the `<PlotContainer>` that Giraffe uses in storybook to make this look cleaner?

Screen Shot 2021-04-08 at 11 36 42 AM

TCL735:
Now becomes this with the adjustment 
Screen Shot 2021-04-08 at 11 44 03 AM

PR #529, opened 2021-04-08
feat: allow 'spec' to be accessed by more components

### LINKED ISSUE #524: Legend: re-organize helper functions and spec availability
1. Currently, the helper functions that shape the data for a `<Tooltip>` for each graph type are in separate locations under `giraffe/src/utils` which makes finding all of them quickly impossible. There should a dedicated folder for all legend and tooltip related code in the utils.

1. With the implementation of a Static Legend, we want to avoid having to re-calculate a graph's `spec` in order to properly get the data series associated with a data set. `spec` should be calculated once and shared with child components who need access.

TCL735:
Closes #524 

- moves 'spec' higher up the component hierarchy to allow Static Legend to access it
- gather up all the utils related to legends and tooltips and put them in one place


jrenee42 (approved):

TCL735:
Resolved merge conflicts

Issue #531, opened 2021-04-08
Annotations: Only allow one tooltip to be shown at a time


hoorayimhelping:
### Background
See influxdata/ui#1121

Right now, tooltips are covering each other if annotations are close to each other. We want to only show a single annotation at once.

When there are multiple annotations, the tooltip should hover based on the annotation closest to the mouse

#### Clarification:
At the location of the mouse, if there are more than 1 overlapping annotations (and its hover range), the annotation that is closest to the mouse should get the hovered state and there should be only 1 tooltip hovered at most

PR #533, opened 2021-04-08
fix: filter out the closest annotation to the mouse to show tootip on

ChitlangeSahas:
closes: #531
https://github.com/influxdata/ui/issues/1117

 Tooltips are covering each other when annotations are close to each other. We want to only show a single annotation tooltip at once.

We filter out the annotation to show the tooltip, by calculating what annotation is closest by distance to the mouse pointer.

TODO: bump up giraffe version for publishing
Video:

https://user-images.githubusercontent.com/18511823/114090196-f6889180-986b-11eb-805c-54b67d2095c6.mov

### DIFF
--- a/giraffe/src/utils/annotationData.ts
+++ b/giraffe/src/utils/annotationData.ts
@@ -1,4 +1,5 @@
 import {AnnotationMark, LineHoverDimension, Scale} from '../types'
+import {min} from 'd3-array'
 
 export const getAnnotationsPositions = (
   annotationData: AnnotationMark[],
@@ -33,6 +34,10 @@ const isWithinHoverableArea = (
   return (
     startValue - margin <= hoverPosition && stopValue + margin >= hoverPosition
   )
+}
+
+const distanceToMousePointer = (annotationMark: AnnotationMark, hoverPosition: number) => {
+  return Math.abs(hoverPosition - annotationMark.startValue)
 }
 
 /***********************************************************************************
@@ -75,6 +80,19 @@ export const getAnnotationHoverIndices = (
             ))
         ) {
           result.push(i)
+
+          if (result.length > 1) {
+            const distances = []
+            result.forEach(i => {
+              const dist = distanceToMousePointer(
+                annotationData[i],
+                dimension === 'x' ? hoverX : hoverY
+              )
+              distances.push(dist)
+            })
+            const closestAnnotation = result[distances.indexOf(min(distances))]
+            result = [closestAnnotation]
+          }
         }
         return result
       }, [])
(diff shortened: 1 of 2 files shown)


daviddkkim:
It looks really smooth based on the video

giraffe/src/utils/annotationData.ts
```diff
@@ -35,6 +36,13 @@ const isWithinHoverableArea = (
   )
 }
 
+const distanceToMousePointer = (
+  annotationMark: AnnotationMark,
+  hoverPosition: number
+) => {
+  return Math.abs(hoverPosition - annotationMark.startValue)
```
TCL735, line 43: This is great and works when Annotations are a single line.

This won't work when Annotations are ranges though. I think we can make a quick tweak to allow it to work with ranges.

- hoverPosition is considered a distance of 0 from the annotation range if it is between the annotation.start and annotation.stop.
- otherwise hoverPosition has a distance that is shortest of: the distance between hoverPosition and annotation.start, and the distance between hoverPosition and annotation.stop

TCL735 (changes requested):

giraffe/src/utils/annotationData.ts
TCL735: i typed that very quickly. i meant `startValue` and `stopValue` but hopefully you get the idea :smile:

ChitlangeSahas: Great point Timmy, I always wondered why there was a `startValue` and `stopValue` when they are the same,  but this makes sense. Didn't know Annotation can be a range. I have some homework due in 2 hrs and I'll address this right after that!

hoorayimhelping:
>you beautiful human being!

daviddkkim (approved):
interaction wise, this looks great

giraffe/src/utils/annotationData.ts
```diff
@@ -35,6 +36,34 @@ const isWithinHoverableArea = (
...
+) => {
+  // check if the Annotation is a range type
+  if (annotationMark.startValue !== annotationMark.stopValue) {
+    if (
+      annotationMark.startValue <= hoverPosition &&
+      hoverPosition <= annotationMark.stopValue
+    ) {
+      // hover position is in the range, effective distance is 0
+      return 0
+    } else {
```
TCL735, line 51: nitpick: you don't need the `else`. The whole block can be outside because the above `if` has a return statement

ChitlangeSahas: Fixed 👍

TCL735 (approved):

hoorayimhelping (approved):

PR #534, opened 2021-04-09
fix: pass <Plot> children down

TCL735:
Fixes failing e2e tests in ui's pipeline where the version is 2.7.2 or 2.7.3

`<Plot>` children were not being passed down and therefore the existing code in ui ignored everything between `<Plot></Plot>` such as the old way of doing Graph + Single Stat


TCL735:
A force push to add the package.json version bump

hoorayimhelping (approved):

Issue #535, opened 2021-04-12
Annotation Tooltip positioning is miscalculating position


hoorayimhelping:
We have some code in `useToolTipStyle` that calculates the tooltip position using complicated math and there is some bug in it that is causing the issues with the tooltip positioning. 

Acceptance criteria: 

* the tooltip position is correctly over the annotation line and clearly out of the way of the click target

TCL735:
Let's reproduce this in Giraffe, standalone without the `ui`, using a storybook example before we try to change the code.

ChitlangeSahas:
I cant produce same behaviors in UI and Giraffe storybook, they're both different and incorrect. As seen below, the last tooltip in giraffe moves downwards and in the UI it goes to the left. 

<img width="1440" alt="Screen Shot 2021-04-13 at 7 48 18 AM" src="https://user-images.githubusercontent.com/18511823/114572836-b428d100-9c2c-11eb-913f-27e80fd946e1.png">

<img width="1440" alt="Screen Shot 2021-04-13 at 7 48 29 AM" src="https://user-images.githubusercontent.com/18511823/114572846-b854ee80-9c2c-11eb-8425-975832c82bd9.png">

PR #537, opened 2021-04-14
fix: fix tooltip positioning when on the edge of the screen

ChitlangeSahas:
Closes: https://github.com/influxdata/giraffe/issues/535

We had an edge case that we missed when calculating the tooltip position. Previous behavior would calculate the positioning when the annotation is on the edge. This PR fixes it so the calculation is correct and the tooltip position is right. 

Co-authored by: Bucky Schwarz (@hoorayimhelping) 

Screenshot:

Big sized tooltip:

Screen Shot 2021-04-13 at 5 49 53 PM

Medium-sized tooltip:

Screen Shot 2021-04-13 at 6 16 00 PM

Small sized tooltip:

Screen Shot 2021-04-13 at 6 18 00 PM

Annotation away from the edge works fine too:

Screen Shot 2021-04-13 at 6 19 00 PM

### DIFF
--- a/giraffe/src/utils/legend/useTooltipStyle.ts
+++ b/giraffe/src/utils/legend/useTooltipStyle.ts
@@ -147,8 +147,8 @@ export const useAnnotationStyle = (
       return {
         display: 'inline-block',
         position: 'fixed',
-        top: `${clampedY}px`,
         left: `${clampedX}px`,
+        top: `${clampedY}px`,
         zIndex: CLOCKFACE_Z_INDEX + LEAFLET_Z_INDEX + 1,
       }
     }
--- a/giraffe/src/utils/legend/useTooltipStyle.ts
+++ b/giraffe/src/utils/legend/useTooltipStyle.ts
@@ -137,7 +137,7 @@ export const useAnnotationStyle = (
 
       // When the annotation is in the far edge of the screen, the position.left value
       // overrides the width of the tooltip and makes its width smaller than its max-width.
-      // Fix the left value so that the tooltip is at its max width.
+      // Position the left edge of the tooltip such that the tooltip occupies its max width.
       if (
         window.innerWidth - clampedX < ANNOTATION_DEFAULT_MAX_WIDTH &&
         tooltipWidth >= ANNOTATION_DEFAULT_MAX_WIDTH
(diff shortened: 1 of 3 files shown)


TCL735:
Can we apply the same logic to vertical annotation tooltips as well? Let's fix `y` while we still have all this context fresh in our brains.

giraffe/src/utils/legend/useTooltipStyle.ts
```diff
@@ -125,13 +132,23 @@ export const useAnnotationStyle = (
         }
       }
 
-      const clampedX = Math.round(Math.max(dx + x, x))
+      let clampedX = Math.round(x + dx)
       const clampedY = Math.round(dy + y)
 
+      // When the annotation is in the far edge of the screen, the position.left value
+      // overrides the width of the tooltip and makes its width smaller than its max-width.
+      // Fix the left value so that the tooltip is at its max width.
```
hoorayimhelping, line 140:
```suggestion
      // Position the left edge of the tooltip such that the tooltip occupies its max width.
```
pedantic, but reading over this comment this morning (i find that reading stuff the day after i wrote it gives me a much fresher perspective on how it reads to someone else), i think it could be a tad bit clearer

giraffe/src/utils/legend/useTooltipStyle.ts
```diff
@@ -125,13 +132,23 @@ export const useAnnotationStyle = (
...
+      if (
+        window.innerWidth - clampedX < ANNOTATION_DEFAULT_MAX_WIDTH &&
+        tooltipWidth >= ANNOTATION_DEFAULT_MAX_WIDTH
```
hoorayimhelping, line 143:
```&&
        tooltipWidth >= ANNOTATION_DEFAULT_MAX_WIDTH```

confused about this condition, what's it for?

ChitlangeSahas: Without it, small tooltips (that have a width smaller than the max width) were being scooted to the side!

hoorayimhelping:
>Can we apply the same logic to horizontal annotation tooltips as well? Let's fix y while we still have all this context fresh in our brains.

@TCL735 @ChitlangeSahas Let's do this in a follow up ticket. I agree we should fix it while the context of this problem is still fresh, but this is a GA blocking bug we need to fix and I want to get it out ASAP. To my knowledge, nobody has tried annotating a vertical graph yet.

hoorayimhelping (approved):

hoorayimhelping:
@TCL735 @ChitlangeSahas https://github.com/influxdata/giraffe/issues/539

TCL735:
Let it rip!

TCL735 (approved):

Issue #543, opened 2021-04-16
Annotations Tooltip: Break long words so they don't extend out of the tooltip


hoorayimhelping:
Long single-word annotations break out of the tooltip. We'll want to break the words using something like

![image.png](https://images.zenhubusercontent.com/5c33b08d24dda4712ef4ae10/ee663402-097e-443c-97a5-865e6cbe7120)

https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
or
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

### Acceptance Criteria
- long words no longer break out of the tooltips

PR #544, opened 2021-04-16
fix: apply 'break-word' to prevent overflow of annotation tooltip text

ChitlangeSahas:
closes: #543

This fixes an issue where the text is a single very long string without breaks was overflowing out of the div.

fix: add css property:  `overflowWrap: 'break-word'`

Note for the reviewer: I tried making a variable to apply the common style to both the title and description `div` elements but kept getting typescript errors. The following was not working.

```
const textContainerStyle = {
    overflowWrap: 'break-word',
  }

...

 <div style={textContainerStyle}>{data.title}</div>
```

Before vs After:

Screen Shot 2021-04-16 at 1 13 16 PM

Screen Shot 2021-04-16 at 1 13 58 PM

### DIFF
--- a/giraffe/package.json
+++ b/giraffe/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/giraffe",
-  "version": "2.7.6",
+  "version": "2.7.7",
   "main": "dist/index.js",
   "module": "src/index.js",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


hoorayimhelping (approved):
lgtm!!!

i think you've earned yourself this mug

PR #555, opened 2021-04-27
feat: add additional columns to Static Legend for stacked line layer

### LINKED ISSUE #548: Static Legend: stacked line layer
The Static Legend should render the "cumulative" column for stacked line layers.

TCL735:
Closes #548 

Adds additional columns in the Static Legend when the line graph is a stacked line layer:
Screen Shot 2021-04-26 at 8 28 53 PM

### DIFF
--- a/giraffe/src/types/index.ts
+++ b/giraffe/src/types/index.ts
@@ -809,12 +809,12 @@ export interface ColumnGroupMap {
 }
 
 export interface LatestIndexMap {
-  xs: {
-    [columnKey: string]: number
-  }
-  ys: {
-    [columnKey: string]: number
-  }
+  // each Column in a Table contains a single array even when there are multiple series
+  //   for example, a line graph has multiple lines (series), but each Column has only one array
+  // Static Legend needs to know where each line (series) ends, to render the "latest"
+  //   - "latest" is the timestamp with the highest value in each line (series)
+  //   - Keep a map of the highest timestamp's index for each series
+  [columnKey: string]: number
 }
 
 export type LineData = {
--- a/giraffe/src/types/index.ts
+++ b/giraffe/src/types/index.ts
@@ -622,7 +622,7 @@ export interface LineLayerSpec {
   }
   columnGroupMaps: {
     fill: ColumnGroupMap
-    latestValueIndices: LatestValueIndexMap
+    latestIndices: LatestIndexMap
   }
   stackedDomainValueColumn?: NumericColumnData
 }
@@ -808,8 +808,13 @@ export interface ColumnGroupMap {
   mappings: Array<{[columnKey: string]: any}>
 }
 
-export interface LatestValueIndexMap {
-  [columnKey: string]: number
+export interface LatestIndexMap {
+  xs: {
+    [columnKey: string]: number
+  }
+  ys: {
+    [columnKey: string]: number
+  }
 }
 
 export type LineData = {
--- a/giraffe/src/types/index.ts
+++ b/giraffe/src/types/index.ts
@@ -622,6 +622,7 @@ export interface LineLayerSpec {
   }
   columnGroupMaps: {
     fill: ColumnGroupMap
+    latestValueIndices: LatestValueIndexMap
   }
   stackedDomainValueColumn?: NumericColumnData
 }
@@ -805,6 +806,10 @@ export interface ColumnGroupMap {
   // A group ID `i` takes on the values specified by `mappings[i]` for the
   // column keys that specify the grouping
   mappings: Array<{[columnKey: string]: any}>
+}
+
+export interface LatestValueIndexMap {
+  [columnKey: string]: number
 }
 
 export type LineData = {
--- a/giraffe/src/utils/legend/sort.test.ts
+++ b/giraffe/src/utils/legend/sort.test.ts
@@ -1,5 +1,5 @@
 import {getDataSortOrder} from './sort'
-import {getRandomTable} from '../randomTable'
+import {getRandomTable} from '../fixtures/randomTable'
 import {lineTransform} from '../../transforms/line'
 import {NINETEEN_EIGHTY_FOUR} from '../../constants/colorSchemes'
 import {DomainLabel} from '../../types'
--- a/giraffe/src/utils/legend/sort.test.ts
+++ b/giraffe/src/utils/legend/sort.test.ts
@@ -30,9 +30,7 @@ describe('getDataSortOrder', () => {
       lineOption
     )
 
-    const latestIndices = Object.values(
-      lineSpec.columnGroupMaps.latestIndices.ys
-    )
+    const latestIndices = Object.values(lineSpec.columnGroupMaps.latestIndices)
     const sortOrder = getDataSortOrder(
       lineSpec.lineData,
       latestIndices,
@@ -54,9 +52,7 @@ describe('getDataSortOrder', () => {
     )
     const {stackedDomainValueColumn} = lineSpec
 
-    const latestIndices = Object.values(
-      lineSpec.columnGroupMaps.latestIndices.ys
-    )
+    const latestIndices = Object.values(lineSpec.columnGroupMaps.latestIndices)
     const sortOrder = getDataSortOrder(
       lineSpec.lineData,
       latestIndices,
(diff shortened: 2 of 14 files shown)


giraffe/src/utils/legend/sort.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  const fillColKeys = ['cpu', 'host', 'machine']
+  const table = getRandomTable(
+    maxValue,
+    true,
+    numberOfRecords,
+    recordsPerLine,
+    fillColKeys
+  )
+
+  it('overlaid line graphs should be unsorted', () => {
```
hoorayimhelping, line 22: it's a little more work, but can you try to convert these into more naturally flowing language? the `it` of the method call is supposed to be read as part of the sentence:

>it overlaid line graphs should be unsorted

might be converted to:

>it leaves overlaid line graphs unsorted

```ts
it('leaves overlaid line graphs unsorted', () => {}
```

it's kind of nitpicky, but it's rspec / jasmine convention, and writing them this way shows up when printing the specs in a report. if you run `jest` with `--verbose` you can see what I'm talking about. Let's just compare two random test suites:

```
 PASS  src/utils/isFiniteNumber.test.ts
  isFiniteNumber
    ✓ the value created from the Number constructor is a finite number (29 ms)
    ✓ the value created from the Number function is a finite number (3 ms)
    ✓ a literal number and constants are finite numbers
    ✓ Infinity and negative Infinity are not finite numbers (1 ms)
    ✓ the value NaN is not a finite number
    ✓ anything else is not a finite number (1 ms)

 PASS  src/utils/replace.test.ts
  replace
    ✓ replaces only the first match of the second argument string with the third argument
    ✓ handles the second argument RegEx with correct replacement using the third argument
    returns the first argument when given less than 3 arguments
      ✓ returns "undefined" as a string when first argument is undefined (1 ms)
      ✓ returns "null" as a string when first argument is null
      ✓ returns the first argument as a string (1 ms)
```

like I said, minor thing, but look how `replace` flows like a sentence, but `isFinite` is kind of disjointed. if all of our tests flow more naturally, we can actually print out our tests in human-readable language, which might be very helpful for not much more effort.

giraffe/src/utils/legend/sort.ts
```diff
@@ -0,0 +1,30 @@
...
+): number[] => {
+  if (!position || position === 'overlaid') {
+    return rowIndices
+  }
+
+  const domainLabelName = domainLabel || DomainLabel.Y
+
+  const dataMap = {}
+  const measurementValues = Object.keys(lineData).reduce(
+    (accumulator, id) => accumulator.concat(lineData[id][domainLabelName]),
```
hoorayimhelping, line 17: what are we actually accumulating? seems like `lineData` ? hinting what this is makes it much easier to follow what we're reducing down.

giraffe/src/utils/legend/sort.ts
```diff
@@ -0,0 +1,30 @@
...
+    []
+  )
+  const sortable = []
+  rowIndices.forEach(hoveredRowIndex => {
+    if (!dataMap[measurementValues[hoveredRowIndex]]) {
```
hoorayimhelping, line 22: what does `measurementValues[hoveredRowIndex]` represent or mean? it might make this function easier to follow if that gets stored into a variable. It'll contextually what `dataMap` is actually mapping

giraffe/src/utils/randomTable.ts
```diff
@@ -173,3 +116,61 @@ export const getRandomTable = memoizeOne(
...
+    .addColumn('_value', 'system', 'number', VALUE_COL)
+
+  if (plotType === LayerTypes.Scatter) {
+    return table
+      .addColumn(POINT_KEY, 'string', 'string', DISK_COL)
+      .addColumn('__symbol', 'string', 'string', SYMBOL_COL)
+      .addColumn(HOST_KEY, 'string', 'string', HOST_COL)
+  }
+  return table.addColumn(COLUMN_KEY, 'string', 'string', CPU_COL)
+}
```
hoorayimhelping, line 176: I'm not sure this belongs in `utils/` which I think of as giraffe's model. This is more of a fixture or helper function, right? We won't every use this inside of giraffe, it's only used to make testing things easier right? If that's the case, it implies it should live somewhere else

hoorayimhelping (commented):
So I don't have any blocking feedback, per se, but there are some points I'd like you to consider, if it's not too much trouble.

giraffe/src/utils/legend/sort.test.ts
TCL735: Oh yeah, definitely. I did that very quickly. Will change the language.

giraffe/src/utils/randomTable.ts
TCL735: It probably doesn't. My reasoning here is that it is useful to open source users who just want some sample data. Having this util is a way of giving that capability to the users.

But on second thought, this can live in fixtures folder and still achieve the above goal. I will move it back into the fixtures folder.

giraffe/src/utils/legend/sort.ts
TCL735: Yep, good call. I think originally I just couldn't come up with good names because I had thought they could be both strings or numbers possibly. But in thinking about how this is used, these are clearly numbers only.

Changed a bunch of names to `numericalValues`, `rowIndex`, `numberMap`, `combinedNumericalValues`. I think it makes the whole function a lot more readable and easier to reason about.

giraffe/src/utils/legend/sort.ts
TCL735: Name changed. See the other comment below.

jrenee42 (changes requested):
please clarify a few things (refactor out some methods for clarity); besides that lt looks good.

hoorayimhelping (approved):
works for me :+1:

giraffe/src/types/index.ts
```diff
@@ -806,6 +807,15 @@ export interface ColumnGroupMap {
   mappings: Array<{[columnKey: string]: any}>
 }
 
+export interface LatestIndexMap {
```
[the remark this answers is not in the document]
TCL735: I added comments to explain what this is:
  `// each Column in a Table contains a single array even when there are multiple series`
  `//   for example, a line graph has multiple lines (series), but each Column has only one array`
  `// Static Legend needs to know where each line (series) ends, to render the "latest"`
  `//   - "latest" is the timestamp with the highest value in each line (series)`
  `//   - Keep a map of the highest timestamp's index for each series`

giraffe/src/utils/legend/staticLegend.ts
```diff
@@ -49,9 +60,42 @@ export const convertLineSpec = (
...
+    additionalColumns.push({
+      key: valueColumnKey,
+      name: STACKED_LINE_CUMULATIVE,
+      type: spec.table.getColumnType(valueColumnKey),
+      colors,
+      values: sortOrder.map(index =>
+        valueFormatter(stackedDomainValues[index])
+      ),
+    })
+
```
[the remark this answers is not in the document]
TCL735: Stacked line graphs are line graphs that progressively add the previous line to the current line to determine the value. The line at the very top of the graph would be the sum of all the lines including itself.

With the above knowledge, the code is actually quite simple. `additionalColumns` are the extra "columns" (like "Latest _value") that stacked line graphs must have. What are they?

- `cumulative` the total value of the line plus all lines below it. each line has its own `_value` which is why a `cumulative` is also required
- `lines` is the count of how many lines are included for this line's `cumulative`. The line with the highest `lines` is the sum of all the other lines and itself. This column is especially important when negative values are present, because the highest line on the graph is not necessarily the line with the most `lines`.


See the Storybook examples to get a feel for what these columns mean.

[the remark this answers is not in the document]
TCL735: Regarding denseness of the code, that is caused by having a lot of necessary options that must be passed to a function. But rest assured, there isn't anything tricky in the code. `Prettier` is very liberal with spacing. Once you get used to reading the code looking at the indentation, there really isn't much going on in this block.

- create a variable
- make a column by pushing to `addtionalColumns`
- create a variable
- update a sort order
- make another column by pushing to `additionalColumns`

PR #556, opened 2021-04-27
feat: clamp hoverX interaction handler to nearest rounded timestamp

hoorayimhelping:
closes influxdata/ui#1265

Timestamps in the graph domain have their dates rounded off based on the size of the domain (so instead of 12:25:33:333 -> 12:25:30). Previously, using the raw `valueX` value from the hover interaction would get the raw timestamp at the point where the mouse is hovering. We now find the nearest rounded timestamp in the domain and set the annotation's time to that.

### DIFF
--- a/giraffe/src/components/SizedPlot.tsx
+++ b/giraffe/src/components/SizedPlot.tsx
@@ -20,11 +20,15 @@ import {
   SpecTypes,
 } from '../types'
 
+import {AnnotationLayer} from './AnnotationLayer'
 import {SingleStatLayer} from './SingleStatLayer'
 import {LineLayer} from './LineLayer'
 import {BandLayer} from './BandLayer'
 import {ScatterLayer} from './ScatterLayer'
 import {RectLayer} from './RectLayer'
+import {MosaicLayer} from './MosaicLayer'
+import GeoLayer from './GeoLayer'
+
 import {Brush} from './Brush'
 import {rangeToDomain} from '../utils/brush'
 import {PlotEnv} from '../utils/PlotEnv'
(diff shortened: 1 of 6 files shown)


giraffe/src/utils/nearestTimestamp.ts
```diff
@@ -0,0 +1,25 @@
...
+    return rawValue
+  }
+
+  if (timestamps.length === 1) {
+    return timestamps[0]
+  }
+
+  const midPoint = timestamps.length / 2
+  const firstHalf = timestamps.slice(0, midPoint)
+  const secondHalf = timestamps.slice(midPoint)
```
hoorayimhelping, line 12: trading some memory and the cost of creating an array for code clarity. We only need to create one array of length `timestamps.length / 2` but this code creates two because it's much easier to follow

giraffe/src/components/SizedPlot.tsx
```diff
@@ -87,10 +89,23 @@ export const SizedPlot: FunctionComponent<SizedPlotProps> = ({
...
 
+  const defaultSpec = env.getSpec(0)
+
+  const valueX = env.xScale.invert(hoverEvent.x)
+  let clampedValueX = NaN
+  if (
+    valueX &&
+    (defaultSpec.type === SpecTypes.Band || defaultSpec.type === SpecTypes.Line)
+  ) {
+    const timestamps = defaultSpec?.lineData[0]?.xs ?? []
```
TCL735, line 100: The layer config will have "x" and "y" properties which are column keys, and of which one of them will be "_time" (usually). Shouldn't the timestamps be based finding out which dimension is "_time"?

Not a deal-breaker since this is only needed to find a clamped value. And currently annotations are supported only when x-axis is time.

TCL735 (approved):

giraffe/src/components/SizedPlot.tsx
hoorayimhelping: this should work on any rounded numerical value, not just timestamps. I'm not sure what is better - explicitly looking for _time, since that what this seems the most relevant to, or using any x axis value, since that's how giraffe handles it?

TCL735: Or find clamped in both dimensions and let the consumer decide if they need it? i dunno. i'm okay with just one dimension for now.

PR #562, opened 2021-04-29
Revert "fix: skip undefined, null, and NaN values during band alignment because mathematical operations cannot be peformed on them"

TCL735:
Reverts influxdata/giraffe#559


giraffe/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/giraffe",
-  "version": "2.8.1",
+  "version": "2.8.0",
```
hoorayimhelping, line 3: need to move this to 2.8.2, right?

hoorayimhelping (approved):

PR #563, opened 2021-04-29
fix: add optional chaining to protect against null from getSpec

TCL735:
The recent re-organization of the `Plot` and `SizedPlot` with new components added in between has brought to light that we must be very careful with changes, as this is the entry point into the heart of the code.

`env.getSpec` can actually return `null` even though it may not happen everywhere. It's still better to guard against this.


hoorayimhelping (approved):
nice catch, thanks!

PR #566, opened 2021-05-04
feat: update Static Legend interface

TCL735:
In support of https://github.com/influxdata/ui/issues/806

- A consuming app may need to adjust `margin` and `padding` around the Static Legend Box, allow this.
- Other styles can be added without further interface changes.
- Does not affect commonly-used strategically-important styles previously defined in the interface.


hoorayimhelping (approved):
do you need to bump package.json?

TCL735:
> do you need to bump package.json?

Not yet. I still need the Scrollbars option and the code for Band Plot.

PR #567, opened 2021-05-04
feat: add style prop to PlotContainer

### LINKED ISSUE #532: Giraffe storybook: PlotContainer should have options
Currently, the `<PlotContainer>` helper component that wraps all `<Plot>`s in the storybook stories use the same base size, padding, and margins.

It would be helpful if `<PlotContainer>` were a little bit more configurable based on the story that we want to write. For example, Annotations have tooltips which require a little bit more space (margin) at the top to properly show a full tooltip that is not cut off. There was a recent change that adjusted the margin for all stories to accommodate Annotations. Ideally, only Annotations' stories should have been adjusted. Having a configurable `<PlotContainer>` will allow this to happen without affecting any other stories.

TCL735:
Closes #532 

Adds a `style` prop to `PlotContainer` to allow for specifying CSS Properties such as margin for the `PlotContainer`

### DIFF
--- a/stories/src/annotation.stories.tsx
+++ b/stories/src/annotation.stories.tsx
@@ -139,7 +139,13 @@ storiesOf('Annotations', module)
       layers,
     }
     return (
-      <PlotContainer>
+      <PlotContainer
+        style={{
+          width: 'calc(100vw - 100px)',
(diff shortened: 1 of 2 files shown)


zoesteinkamp (approved):
yep yep more styling

PR #570, opened 2021-05-05
fix: this is what broke everything

zoesteinkamp:
fix: force coercion of falsy values to 0

guards against https://github.com/d3/d3-scale/releases/tag/v3.2.4

Issue: on april 28th 2021, we found an issue when we upgraded multiple giraffe libraries. This resulted in our graphs intermittently coming back empty/blank. This was because we had originally believed the null values would return 0, as d3-scale did originally. But by upgrading this broke and we were receiving NaN values. These values made it so our graphs could not be displayed properly but only broke on graphs with undefined values.

### DIFF
--- a/giraffe/src/utils/lineData.ts
+++ b/giraffe/src/utils/lineData.ts
@@ -41,8 +41,8 @@ export const simplifyLineData = (
 
   for (const [k, {xs, ys, fill}] of Object.entries(lineData)) {
     const [simplifedXs, simplifiedYs] = simplify(
-      xs.map(x => xScale(x)),
-      ys.map(y => yScale(y)),
+      xs.map(x => xScale(x || 0)),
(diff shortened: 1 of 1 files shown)


hoorayimhelping (changes requested):
I got a chuckle out of the cheeky commit message, but let's think of posterity on this one. Describe with a little more detail what this fix is - it's funny now, I'll give you that, but in six months, I doubt we'll remember what 'everything' included when we 'broke everything'

Put it in your own style, but if it was me, I'd make this message something like:
```sh
fix: force coercion of falsy values to 0

guards against https://github.com/d3/d3-scale/releases/tag/v3.2.4
```

it gives a quick tldr in the first line, and then provides context by linking to the release in the detail part of the message

hoorayimhelping (approved):

PR #578, opened 2021-05-10
Refactor/geo

### LINKED ISSUE #573: Maps: Refactor Geo Code
Refactoring the Maps Geo Code in giraffe:
- [x] Remove the Abstract Geo class and instead implement the GeoTable into Pivoted and Native Geo Table classes.
- [x] Break down large components into smaller components.
- [x] Move certain functionality from components to utility files/folders
- [x] Clean unnecessary code/implementations
- [ ] Find any error use cases and fail gracefully (for example: in getLatLon function)

Palakp41:
Closes #573 

- [x] Remove the Abstract Geo class and instead implement the GeoTable into Pivoted and Native Geo Table classes.
- [x] Break down large components into smaller components.
- [x] Move certain functionality from components to utility files/folders
- [x] Clean unnecessary code/implementations


giraffe/src/components/geo/LayerSwitcher.tsx
```diff
@@ -0,0 +1,81 @@
...
+  stylingConfig: Partial<Config>
+}
+
+export const LayerSwitcher: FunctionComponent<Props> = ({
+  layer,
+  preprocessedTable,
+  index,
+  stylingConfig,
+}) => {
+  switch (layer.type) {
```
hoorayimhelping, line 33: question: is a `default` necessary here? if it can't fail here, a comment saying our code can't reach this is a good idea.

since this is a library, we may need to guard against this anyway in the future

giraffe/src/components/geo/processing/mixins.ts
```diff
@@ -0,0 +1,57 @@
...
+  TIME_COLUMN,
+} from './tableProcessing'
+
+export const getLatLonMixin = function(index: number): LatLon {
+  if (this.coordinateEncoding === CoordinateEncoding.FIELDS) {
+    return {
+      lon: this.getValue(index, LON_COLUMN),
+      lat: this.getValue(index, LAT_COLUMN),
+    }
+  } else {
```
hoorayimhelping, line 18: ✂️ this `else`

giraffe/src/components/geo/processing/mixins.ts
```diff
@@ -0,0 +1,57 @@
...
+    const cellId = this.getS2CellID(index)
+    if (cellId === null || cellId.length > 16) {
+      return null
+    }
+    const fixed =
+      BigInt('0x' + cellId) * PRECISION_TRIMMING_TABLE[16 - cellId.length]
+    const latLng = S2.idToLatLng(fixed.toString()) // S2LatLng.fromInteger(fixed)
```
hoorayimhelping, line 25: remove the comment at the end of the line 

`// S2LatLng.fromInteger(fixed)`

giraffe/src/components/geo/processing/mixins.ts
```diff
@@ -0,0 +1,57 @@
...
+      lat: latLng.lat,
+    }
+  }
+}
+
+export const getTimeStringMixin = function(index: number): string {
+  const timeValue = this.getValue(index, TIME_COLUMN)
+  if (timeValue) {
+    return timestampToString(timeValue)
+  } else {
```
hoorayimhelping, line 37:
```suggestion
  }
```

giraffe/src/components/geo/processing/mixins.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const getLatLonMixin = function(index: number): LatLon {
+  if (this.coordinateEncoding === CoordinateEncoding.FIELDS) {
+    return {
+      lon: this.getValue(index, LON_COLUMN),
+      lat: this.getValue(index, LAT_COLUMN),
+    }
+  } else {
+    const cellId = this.getS2CellID(index)
+    if (cellId === null || cellId.length > 16) {
```
hoorayimhelping, line 20: let's pull this `16` into a variable or leave a comment explaining why 16 is important

giraffe/src/components/geo/processing/GeoTable.ts
```diff
@@ -3,6 +3,11 @@ export interface LatLon {
   lat: number
 }
```
hoorayimhelping, line 4: `LatLon` => `Coordinate`

giraffe/src/components/geo/processing/PivotedGeoTable.ts
```diff
@@ -76,6 +77,12 @@ export class PivotedGeoTable extends AbstractGeoTable {
       this.truncated = false
     }
   }
+  mapTracks<T, U>(
```
hoorayimhelping, line 80: it might make sense to move `mapTracks` function out of `GeoTable` interface, and just have `NativeGeoTable` implement it.

giraffe/src/components/geo/processing/mixins.ts
```diff
@@ -0,0 +1,57 @@
...
+      return null
+    }
+    const fixed =
+      BigInt('0x' + cellId) * PRECISION_TRIMMING_TABLE[16 - cellId.length]
```
hoorayimhelping, line 24: this damn `16` again :D

giraffe/src/components/geo/processing/mixins.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+}
+
+const timestampToString = timeValue => new Date(timeValue).toLocaleString()
+
+const PRECISION_TRIMMING_TABLE = [BigInt(1)]
+for (let i = 1; i < 17; i++) {
+  PRECISION_TRIMMING_TABLE[i] = PRECISION_TRIMMING_TABLE[i - 1] * BigInt(16)
+}
```
hoorayimhelping, line 57: let's move 
```const PRECISION_TRIMMING_TABLE = [BigInt(1)]``` 
to the top of the file and move 

```
for (let i = 1; i < 17; i++) {
  PRECISION_TRIMMING_TABLE[i] = PRECISION_TRIMMING_TABLE[i - 1] * BigInt(16)
}
```

into a function that you can call in `getLatLonMixin`

File: giraffe/src/utils/geo.ts [hunk 10]
```diff
@@ -0,0 +1,205 @@
...
+    colorValue,
+    colorDimension
+  )
+  if (colorInfo) {
+    result.push(colorInfo)
+  }
+  return result
+}
+
+const getClusterValues = (cluster, result: number[]): GeoPointMapViewLayer => {
```
hoorayimhelping, line 128: let's add a comment calling out the fact that `result is overwritten`

giraffe/src/utils/geo.ts
```diff
@@ -0,0 +1,205 @@
...
+  const aggregation =
+    properties.clusterAggregationFunction || ClusterAggregation.max
+  switch (aggregation) {
+    case ClusterAggregation.max:
+      return {value: Math.max(...values), properties}
+    case ClusterAggregation.min:
+      return {value: Math.min(...values), properties}
+    case ClusterAggregation.mean:
+      return {
+        value: values.reduce((a, x) => a + x, 0) / values.length,
```
hoorayimhelping, line 157: if `values` is empty, this crashes because of divide by zero, so we should check for it being zero and handle it.

giraffe/src/components/geo/processing/PivotedGeoTable.ts
Palakp41: After some investigation, I realize this is needed in PivotedGeoTable. If we pass in the values for lat/lon as fields it should still render the map.

Currently, all the layers are using `GeoTable` type for props and depending on the native/pivoted that function is being executed. When we implement the TrackMaps functionality we will need to update/fix this code. Right now, it is hard coded to use the `lat` and `lon` columns for Native Geo Table. If the data doesn't include those two columns, the map fails horribly. So, for now I am leaving this as is, but creating an issue that will be put into later enhancement epics for track maps.

Palakp41: Issue created to track this: https://app.zenhub.com/workspaces/applicationui-5e3b05772922e316b3a210a4/issues/influxdata/giraffe/584

hoorayimhelping (approved):

hoorayimhelping:
Nice work!

Issue #343, opened 2020-10-29
Customization / clean up of units


hoorayimhelping:

121watts:
@alexpaxton and @russorat please go over this.

zoesteinkamp:
This issue is a proof of concept for some of these ideas in giraffe
https://github.com/influxdata/giraffe/pull/580

If we want to see this work fully it will take more time than we are currently willing to dedicate to it.

PR #580, opened 2021-05-10
feat: adding units to the gauge

zoesteinkamp:
This is the first step for issue 19880, this is the giraffe addition that needs to now be added to UI. Which is when i will close that issue. 

But it does close this issue #343. Only thing to note is this does only add this functionality to gauge. If we want to add it to other graph types we need to create a separate issue for that.

### DIFF
--- a/giraffe/README.md
+++ b/giraffe/README.md
@@ -717,6 +717,8 @@ TableGraphLayerConfig uses the `fluxResponse` property from `config` as the data
 
   - **tickSuffix**: _string. Optional._ The text that appears after each tick label. Use an empty string if no text is preferred.
 
+  - **gaugeUnit**: _select. Optional._ It allows the user to pick a unit of byte, usd, or time. Automatically set to none. This will change the tick prefix to change to the preset units. If Gauge units is selected lineCount is overwritten and will not respond to inputs.
+
   - **decimalPlaces**: _Object. Optional._
 
     - **isEnforced**: _boolean. Optional. Defaults to false when not included._ Indicates whether the number of decimal places ("**digits**") will be enforced. When **isEnforced** is falsy or omitted, **digits** will be locked to 2 for stat values with a decimal and 0 for stat values that are integers, and the **digits** option will be ignored.


TCL735 (changes requested):
- Selecting 'time' or 'usd' causes the gauge to render an "undefined" below the right side of the gauge.

- Selecting any unit (except 'none') cause the Gauge Lines option not to work. The gauge no longer renders the correct number of lines.

TCL735:
It seems like the pre-defined units have a fixed scale, and the scale is exponential. For example, time goes from seconds all the way to 30 days within the same gauge, with each evenly spaced section jumping from 1 minute -> 1 hour -> 12 hours -> 24 hours -> 30 days.

Wouldn't this cause the gauge to stay in a fixed position if my data set has only values between, say, 0 minutes to 100 minutes? The gauge would hardly move at all and it would be hard to distinguish between the low end versus the high end values.

And what happens if my data set goes beyond 30 days? Seems like the gauge would not be able to render it.

giraffe/src/components/Gauge.tsx
```diff
@@ -295,55 +313,121 @@ const drawGaugeLabels = (
...
-  for (let i = 0; i <= lineCount; i++) {
-    labelRadius = radius + gradientThickness + 23
-    if (lineCount === 1 && i === 1) {
-      ctx.textAlign = 'left'
-    } else if (lineCount % 2 === 0 && lineCount / i === 2) {
-      ctx.textAlign = 'center'
-    } else if (i / (lineCount + 1) >= 0.5) {
-      ctx.textAlign = 'left'
+  if (gaugeUnit.toString() === 'bytes') {
+    const labels = ['0b', '1024Kb', '1024Mb', '1024Gb', '1024Tb', '1024Pb']
```
TCL735, line 333: I don't think this is the most useful way to do units.

giraffe/src/components/Gauge.tsx
```diff
@@ -295,55 +313,121 @@ const drawGaugeLabels = (
...
+      ctx.rotate(-startDegree)
+      ctx.fillText(labels[i], 0, 0)
+      ctx.rotate(startDegree)
+      ctx.rotate(i * arcIncrement)
+      ctx.translate(-labelRadius, 0)
+      ctx.rotate(i * -arcIncrement)
+      ctx.rotate(-startDegree)
+    }
+  } else if (gaugeUnit.toString() === 'USD') {
+    const labels = ['0', '100', '1000', '1000m', '1000b', '1000t']
```
TCL735, line 375: I don't think this is the most useful way to do units.

giraffe/src/components/Gauge.tsx
```diff
@@ -295,55 +313,121 @@ const drawGaugeLabels = (
...
+      ctx.rotate(-startDegree)
+      ctx.fillText(labels[i], 0, 0)
+      ctx.rotate(startDegree)
+      ctx.rotate(i * arcIncrement)
+      ctx.translate(-labelRadius, 0)
+      ctx.rotate(i * -arcIncrement)
+      ctx.rotate(-startDegree)
+    }
+  } else if (gaugeUnit.toString() === 'time') {
+    const labels = ['0', '60s', '60m', '12h', '24h', '30d']
```
TCL735, line 354: I don't think this is the most useful way to do units.

TCL735 (changes requested):

PR #583, opened 2021-05-12
fix: add null check in PlotEnv

TCL735:
Not all layers will create a spec, such as `RawFluxDataTable`, `Gauge`, `Geo`, `Custom`, `SingleStat`, and `Table`.
See https://github.com/influxdata/giraffe/blob/master/giraffe/src/utils/PlotEnv.ts#L375-L381

As such, guard against accessing a property on null.


hoorayimhelping (approved):
nice find!

PR #589, opened 2021-05-25
feat: drawing range annotations

jrenee42:
for giraffe#585

this PR is all about drawing annotations.

users of this library (ui, in our case) has its own listeners for graph interaction to add annotations

added a new interaction handler; so that when brushing on the x axis, a callback other than zoom can be invoked

also: differentiating now properly between drags and single clicks.

### DIFF
--- a/giraffe/src/components/AnnotationLine.tsx
+++ b/giraffe/src/components/AnnotationLine.tsx
@@ -6,6 +6,7 @@ import styles from './AnnotationLine.scss'
 interface AnnotationLineProps {
   dimension: AnnotationDimension
   color: string
+  secondaryColor?: string
   strokeWidth: number
   startValue: number
   stopValue: number
@@ -20,13 +21,23 @@ const PIN_TRIANGLE_HEIGHT = 8
 const PIN_TRIANGLE_WIDTH = 6
 
 export const AnnotationLine: FunctionComponent<AnnotationLineProps> = props => {
-  const {dimension, color, strokeWidth, startValue, length, pin} = props
+  const {
+    dimension,
+    color,
+    secondaryColor,
+    strokeWidth,
+    startValue,
+    stopValue,
+    length,
+    pin,
+  } = props
 
   // This prevents blurry sub-pixel rendering as well as clipped lines
   // If the line is at the edge of the canvas the stroke will be half obscured
   // because the stroke is centered on the line. Giving a minimum value
   // prevents the line from being clipped
   const clampedStart = Math.max(1, Math.round(startValue))
+  const clampedEnd = Math.max(1, Math.round(stopValue))
 
   if (dimension === 'y') {
     return (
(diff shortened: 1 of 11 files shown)


giraffe/src/components/SizedPlot.tsx
```diff
@@ -91,18 +83,38 @@ export const SizedPlot: FunctionComponent<SizedPlotProps> = ({
 
   const defaultSpec = env.getSpec(0)
 
-  const valueX = env.xScale.invert(hoverEvent.x)
-  let clampedValueX = NaN
+  // get me the x value (which might be a timestamp) of a particular
+  // point on the x axis
+  const getValueX = (point, snapToDataPoint = true) => {
```
hoorayimhelping, line 88: boolean arguments are usually a smell, and I think this is a case of premature abstraction. To make this function work, we basically have to call it in three different configurations in the three places its called. In my opinion, it just obfuscates the meaning of code and creates a needless function, since we're basically replicating the three lines the function replaced.

We could get the same meaning and the same relationship in the same number of lines, without the need of an extra function to refer to, by just having a couple of first class function calls.

The function isn't needed in this call at all, since the value returned from `env.xScale.invert` is used:
```ts
const valueX = getValueX(hoverEvent.x, false)
const clampedValueX = getNearestTimeStamp(valueX)
```
could then become:
```ts
const valueX = env.xScale.invert(hoverEvent.x)
const clampedValueX = getNearestTimeStamp(valueX)
```
It's easier to read - you don't have to go to a function and see that `env.xScale.invert` is called, you just see it called here.

---

This code
```ts
const beginning = getValueX(xRange[0], true)
const end = getValueX(xRange[1], true)
```
could be changed to something like this:

```ts
const beginning =  getNearestTimeStamp(env.xScale.invert(xRange[0]))
const end = getNearestTimeStamp(getValueX(xRange[1]))
```

This code is the same number of lines, and communicates the same thing, but doesn't require a second function wrapped around `invert`.

I think this function is one of those cases where being prematurely DRY actually makes the code harder to comprehend.

giraffe/src/utils/useDragEvent.ts
```diff
@@ -6,13 +6,64 @@ import {useForceUpdate} from './useForceUpdate'
...
+ */
+
 export interface DragEvent {
   type: 'drag' | 'dragend'
   initialX: number
   initialY: number
   direction: 'x' | 'y' | null
   x: number
   y: number
+  mouseState: 'mouseUpHappened' | 'mouseDownHappened' | null
```
hoorayimhelping, line 65: let's call this something like `mouseClickState` since the `mouseState` might be assumed to have x and y coordinates

hoorayimhelping (commented):
I don't have any blocking feedback on this, but I do have some strong opinions on that helper function and premature abstraction. Let me know what you think

giraffe/src/components/AnnotationLine.tsx
```diff
@@ -69,59 +106,116 @@ export const AnnotationLine: FunctionComponent<AnnotationLineProps> = props => {
...
+          ${clampedEnd + pixelMargin}, ${RANGE_HEIGHT}
+          ${clampedStart - pixelMargin}, ${RANGE_HEIGHT}`,
+      fill: color,
+      id: props.id,
+      style: {cursor: 'pointer'},
+    })
+  }
+
+  // this is the overlay that goes over the whole range; with 10% opacity
+  // do not make this click to edit, b/c then cannot have overlapping point annotations
```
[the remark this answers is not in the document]
jrenee42: but yet we need to state the consequences of making the overlay click to edit.

do you a suggestion of what to put in instead?

giraffe/src/components/AnnotationLine.tsx
```diff
@@ -69,59 +106,116 @@ export const AnnotationLine: FunctionComponent<AnnotationLineProps> = props => {
...
+          ${clampedEnd}, ${length}
+          ${clampedEnd}, ${RANGE_HEIGHT}
+          ${clampedStart}, ${RANGE_HEIGHT}`,
+      fill: color,
+      opacity: 0.1,
+    })
+  }
+
+  const makePin = (pinType?: string, timeSignature?: number) => {
+    if (!timeSignature) {
```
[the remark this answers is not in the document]
jrenee42: there was originally pins at the beginning and end of the ranged annotations, hence why this was refactored out.  I can remove that part; as will now always be the same.

giraffe/src/components/SizedPlot.tsx
jrenee42: whereas the 'invert' function is what is confusing to me; hence why I renamed it.  and you did include 'getValueX' in the second line of what you suggested (assume that was a mistake....)

giraffe/src/components/AnnotationLine.tsx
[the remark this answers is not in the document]
jrenee42: ...wait, I see you offered a suggestion.  will edit it :)

giraffe/src/utils/useDragEvent.ts
jrenee42: but it's not a click.... when there is dragging.  'mouseActionState' works tho (it's an action that the mouse does, but not necessarily a click)

giraffe/src/components/SizedPlot.tsx
jrenee42: i understand wanting less lines of code, but the 'invert' makes no sense to me; it isn't explained/documented anywhere.
......pondering.....

jrenee42: i propose that getNearestTimeStamp be renamed to getNearestDataPoint
which is more accurate
and it takes the actual graph point and does the env.xScale.invert

because it is a smell for the method to have submethods called within its arguments.
that way, it will be clearer what is going on,
and the 'getValueX' wrapper will be removed

giraffe/src/utils/useDragEvent.ts
hoorayimhelping: makes sense, carry on.

giraffe/src/components/SizedPlot.tsx
hoorayimhelping: fair enough! thanks for listening

hoorayimhelping (approved):
hold on to your butts!

PR #591, opened 2021-05-25
fix: refactor tooltip and static legend value sorting to use common functionality

### LINKED ISSUE #590: Static Legend: colors are misaligned
Colors in the static legend do not match the graph nor the tooltip.

Here is an example:
image (3)

In addition: values at the edge of the graph are conflicting--the tooltip and static legend do not agree. Here are two examples:
Screen Shot 2021-06-01 at 7 28 02 PM
Screen Shot 2021-06-01 at 7 28 30 PM

TCL735:
Closes #590 

- fill columns (not colors, which were always correct) have been fixed to align properly and show correct values
- tooltip and static legend are now:
  - sorted by the value column in descending order
  - use the same functions and flow in calculating the data for the their respective columns
  - always in agreement at the edge of the graph as a result of the above
- band plot's underlying logic for legend has been cleaned up by using better names and less confusing functions
- unit tests have been added related to the above

Screen Shot 2021-05-25 at 4 52 21 PM
Screen Shot 2021-05-25 at 4 52 38 PM
Screen Shot 2021-05-25 at 4 52 32 PM


TCL735:
A couple of minor tweaks.

Ready for review now.

Palakp41 (approved):
LGTM!

giraffe/src/utils/legend/staticLegend.ts
```diff
@@ -138,16 +138,17 @@ export const convertLineSpec = (
...
 
     return {
       key,
       name: key,
       type: spec.table.getColumnType(key),
-      values: column,
+      values: fillColumn,
       colors,
     }
   })
```
hoorayimhelping, line 154: is this function unit testable? seems like a good candidate for some unit tests.

hoorayimhelping (commented):
what do you think about adding some unit tests?

giraffe/src/utils/legend/staticLegend.ts
TCL735: It is, but an assumption has to be made. We have to assume that line graphs will sort the legend by value.

The reason there hasn't been unit tests for this previously was because line graphs (in overlaid mode) were not sorted in the tooltip (or legend as we call it now). This was quietly changed with the initial implementation of Static Legend with multiple components now consuming the underlying legend component.

I think it's time to codify the assumption and as you suggested, add some unit tests for this.

giraffe/src/transforms/band.ts
```diff
@@ -437,38 +438,28 @@ export const bandTransform = (
...
       //    it will be updated with another loop later,
-      //    because it is faster to walk bandIndexMap once
+      //    because it is faster to walk bandLineMap once
       //    than to search for the indices of many lines
     }
 
     lineData[groupID].xs.push(x)
     lineData[groupID].ys.push(y)
 
-    // remember the latest (most recent) index for each group
```
ShmuelLotman, line 447: Sorry, asking bc i just dont have context into this, so im curious - what is the purpose of this deleted code and why is it being removed?

ShmuelLotman (approved):

giraffe/src/transforms/band.ts
TCL735: That chunk of code was causing the misalignment and inconsistencies between the tooltip and the static legend. It is being replaced further down in the file by a function that will be shared between tooltip and static legend.

PR #593, opened 2021-05-27
chore: enable band annotations, also bump version number up

jrenee42:


giraffe/src/components/SizedPlot.tsx
```diff
@@ -88,9 +88,16 @@ export const SizedPlot: FunctionComponent<SizedPlotProps> = ({
...
+    // need to add the type below in the 'if' statement clause
+
     // now find the nearest data point:
     let nearest = NaN
-    if (valueX && defaultSpec?.type === SpecTypes.Line) {
+    if (
+      valueX &&
+      (defaultSpec.type === SpecTypes.Line ||
+        defaultSpec.type === SpecTypes.Band)
+    ) {
```
genehynson, line 100: could we convert this to a list? Like `let annotationPlots = [ SpecTypes.Line, SpecTypes.Band ]` and then this if statement would just be `if (valueX && defaultSpec.type in annotationPlots) ...` That way this `if` doesn't expand with a bunch of conditions and get hard to read

jrenee42: we could, but then it might take a little longer to process; and this is processing with each graph redraw.
plus; at the moment we only have two and it's going to stay this way for a while.
if it goes above three then a list  would be good.

genehynson: If you use a set or dictionary it is still O(1)! I just think that if we're going to expand this list of accepted types in the future, it would be good to structure it in an expandable way now

TCL735: We need to put the `?` back in like this `defaultSpec?.type` or else this will break on graph types that don't have a spec

TCL735 (changes requested):

giraffe/src/components/SizedPlot.tsx
jrenee42: tldr:  using a Set causes the typescript transpiler/webpack to error; since the equality limits the type:

so; changing the code to:
```
const annotationSpecTypes = new Set()

  annotationSpecTypes.add(SpecTypes.Line)
  annotationSpecTypes.add(SpecTypes.Band)

  const getNearestDataPoint = graphPoint => {
    // this gets the actual value of the point given:
    // (the x value of a particular point on the x axis)
    const valueX = env.xScale.invert(graphPoint)

    // NOTE:  every time a new plot type is added to annotations,
    // need to add the type below in the 'if' statement clause

    // now find the nearest data point:
    let nearest = NaN
    if (valueX && annotationSpecTypes.has(defaultSpec.type)) {
      const timestamps = defaultSpec?.lineData[0]?.xs ?? []
      nearest = nearestTimestamp(timestamps, valueX)
    }
    return nearest
  }
```

yields this error:

```
ERROR in [at-loader] ./src/components/SizedPlot.tsx:102:39
    TS2339: Property 'lineData' does not exist on type 'LayerSpec'.
  Property 'lineData' does not exist on type 'AnnotationLayerSpec'.
```

genehynson: The issue is that one type is a string, the other is a `SpecTypes` enum. I believe we should change the `type` property to be of type `SpecTypes` here: https://github.com/influxdata/giraffe/blob/master/giraffe/src/types/index.ts#L582

`type: 'annotation'` --> `type: SpecTypes.Annotation`

TCL735 (approved):

giraffe/src/components/SizedPlot.tsx
jrenee42: the spectypes already include 'annotation'.  and all the layer configs use the raw strings .  so that is already set.
```

export enum SpecTypes {
  Annotation = 'annotation',
  Line = 'line',
  Band = 'band',
  Scatter = 'scatter',
  Rect = 'rect',
  Mosaic = 'mosaic',
  Table = 'table',
}
```

ie:
```

export interface LineLayerSpec {
  type: 'line' // do not refactor or restrict to SpecTypes.Line
  inputTable: Table
  table: Table // has `FILL` column added
  lineData: LineData
  xDomain: number[]
  yDomain: number[]
  xColumnKey: string
  yColumnKey: string
  xColumnType: ColumnType
  yColumnType: ColumnType
  scales: {
    fill: Scale<number, string>
  }
  columnGroupMaps: {
    fill: ColumnGroupMap
    latestIndices: LatestIndexMap
  }
  stackedDomainValueColumn?: NumericColumnData
}

```

PR #596, opened 2021-05-28
Allow maps latlon/s2_cell_id to be user selected

Palakp41:
In the config that we pass into Geo for render, we have added more properties to handle user selected values for lat/lon or s2 cell id. We have now removed the hardcoded values from codebase used for lat/lon or s2_cell_id.


giraffe/src/components/Geo.tsx
```diff
@@ -28,7 +28,17 @@ const Geo: FunctionComponent<Props> = props => {
...
+    lat,
+    lon,
+    zoom,
+    mapStyle,
+    stylingConfig,
+    allowPanAndZoom,
+    useS2CellID,
+    latLonColumns,
+    s2Column,
+  } = props
```
hoorayimhelping, line 41: nit: alphabetizing these would make seeing changes easier

giraffe/src/components/geo/processing/tableProcessing.ts
```diff
@@ -26,12 +27,26 @@ export const GEO_HASH_COLUMN = 's2_cell_id'
 export const preprocessData = (
```
hoorayimhelping, line 27: can you add some unit tests for this function? basically test that it returns the type of table you expect given all the inputs and outputs?

hoorayimhelping (commented):
>17,968 additions, 0 deletions not shown because the diff is too large. Please use a local Git client to view these changes.

wow. looks good, but what do you think of adding a unit test for `preprocessData`?

giraffe/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/giraffe",
-  "version": "2.10.6",
+  "version": "2.11.0",
```
hoorayimhelping, line 3: this will need to be updated

hoorayimhelping (approved):

PR #601, opened 2021-06-04
feat: disable the hover layers for line and band when legendHide is truthy to avoid confusing the user

TCL735:
Part of https://github.com/influxdata/ui/issues/1456

To avoid confusing the user and leading them to thinking there's a bug, the hover layer should be disabled when `legendHide` is truthy.

When disabled, the cursor is now 'auto' (not a crosshair), and there is no guide wire. 

https://user-images.githubusercontent.com/10736577/120868518-13c69d00-c549-11eb-82a0-05dca0750fd4.mp4

### DIFF
--- a/giraffe/src/types/index.ts
+++ b/giraffe/src/types/index.ts
@@ -689,17 +689,18 @@ export interface RectLayerSpec {
 }
 
 export interface LayerProps {
-  xScale: Scale<number, number>
-  yScale: Scale<number, number>
-  width: number
-  height: number
-  spec: LayerSpec
+  columnFormatter: (colKey: string) => (x: any) => string
+  yColumnType: ColumnType
   config: LayerConfig
   plotConfig: SizedConfig
-  yColumnType: ColumnType
-  columnFormatter: (colKey: string) => (x: any) => string
+  spec: LayerSpec
   hoverX: number | null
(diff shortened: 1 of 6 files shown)


TCL735:
Force push to resolve merge conflict.

Ready for review.

giraffe/src/types/index.ts
```diff
@@ -689,17 +689,18 @@ export interface RectLayerSpec {
...
+  columnFormatter: (colKey: string) => (x: any) => string
+  yColumnType: ColumnType
   config: LayerConfig
   plotConfig: SizedConfig
-  yColumnType: ColumnType
-  columnFormatter: (colKey: string) => (x: any) => string
+  spec: LayerSpec
   hoverX: number | null
   hoverY: number | null
+  legendHide: boolean
```
Palakp41, line 699: ❓ :  Since this is not an optional parameter, would it be a breaking change in ui?

Palakp41 (approved):
Looks good!

giraffe/src/types/index.ts
TCL735: It is not optional but will not break the ui because this is in `LayerProps` which is only an internal type. It does get exported out of index.ts but it has not been exported at the library level.

PR #604, opened 2021-06-08
feat: brush control keys shift

jrenee42:
now only executing interactionaction onXBrush when the shift key is down

### DIFF
--- a/giraffe/package.json
+++ b/giraffe/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/giraffe",
-  "version": "2.13.2",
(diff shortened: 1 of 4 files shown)


giraffe/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/giraffe",
-  "version": "2.13.1",
+  "version": "2.13.2",
```
TCL735, line 3: I suggest bumping this 2.14.0 because we're defining something new here rather than fixing a bug.

TCL735 (changes requested):

PR #606, opened 2021-06-09
Maps tooltip

### LINKED ISSUE #557: Maps: Update data point tool tip to include all tags
Allow more than one field to display on the dropped pin?  No additional fields, but all Tagged values

zoesteinkamp:
closes #557 

This makes it so the tooltips for maps does change with the changing of queries in UI


giraffe/src/components/geo/PointMapLayer.tsx
```diff
@@ -72,6 +74,7 @@ export const PointMapLayer: FunctionComponent<Props> = props => {
         ])
       }}
     />
+    // <Tooltip data={properties} config={stylingConfig} />
```
Palakp41, line 77: Can we remove this commented code?

Palakp41 (changes requested):
I think you need to bump giraffe version so that we can publish and pull latest in ui.

Palakp41 (approved):
Looks good!

PR #607, opened 2021-06-09
chore: adding pretty-quick

jrenee42:
very simple.

not bumping the version number since I'll be doing another pr  for annotations and will bump it then.

### DIFF
--- a/giraffe/package.json
+++ b/giraffe/package.json
@@ -7,8 +7,8 @@
   "scripts": {
     "build": "rm -rf dist && NODE_ENV=production webpack --config webpack.config.js",
     "lint": "eslint '{src,../stories/src}/**/*.{ts,tsx}'",
-    "prettier": "prettier --config ../.prettierrc.json --check '{src,../stories/src}/**/*.{ts,tsx}'",
-    "prettier:fix": "prettier --config ../.prettierrc.json --write '{src,../stories/src}/**/*.{ts,tsx}'",
+    "prettier": "pretty-quick --config ../.prettierrc.json --check '{src,../stories/src}/**/*.{ts,tsx}'",
+    "prettier:fix": "pretty-quick --config ../.prettierrc.json --write '{src,../stories/src}/**/*.{ts,tsx}'",
     "publish": "echo 'To publish giraffe, run ./publish $version'",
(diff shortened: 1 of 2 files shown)


giraffe/package.json
```diff
@@ -107,6 +107,7 @@
     "react-dom": "^16.8.0"
   },
   "dependencies": {
-    "merge-images": "^2.0.0"
+    "merge-images": "^2.0.0",
+    "pretty-quick": "^3.1.0"
```
TCL735, line 111: This should be in `devDependencies` just like `prettier` is.

TCL735 (changes requested):

giraffe/package.json
jrenee42: got it.  manually moved it over.

giraffe/package.json
```diff
@@ -7,8 +7,8 @@
   "scripts": {
     "build": "rm -rf dist && NODE_ENV=production webpack --config webpack.config.js",
     "lint": "eslint '{src,../stories/src}/**/*.{ts,tsx}'",
-    "prettier": "prettier --config ../.prettierrc.json --check '{src,../stories/src}/**/*.{ts,tsx}'",
-    "prettier:fix": "prettier --config ../.prettierrc.json --write '{src,../stories/src}/**/*.{ts,tsx}'",
+    "prettier": "pretty-quick --config ../.prettierrc.json --check '{src,../stories/src}/**/*.{ts,tsx}'",
+    "prettier:fix": "pretty-quick --config ../.prettierrc.json --write '{src,../stories/src}/**/*.{ts,tsx}'",
```
TCL735, line 11: I would like to preserve the original scripts as well. The originals can have new names. Something like `prettier:all` and `prettier:all:fix`

TCL735: This would be the last issue and then this PR will be good to go. Thanks.

TCL735 (changes requested):

TCL735 (approved):

giraffe/package.json
jrenee42: ps:  none of the other projects preserve the original scripts.  so you can add them back if you like. (clockface and ui)

PR #621, opened 2021-06-17
chore: change to shift click

jrenee42:
now the user must hold the shift key down when clicking to create a point annotation.

updated the interaction listener code (added a new one, kept the old one in for backwards compatibility for other users)

also bumped up the version number


zoesteinkamp (approved):
looks good to me

PR #623, opened 2021-06-22
Fix/maps pivoted tooltips

Palakp41:
The `PivotedGeoTable` doesn't show tooltips as it doesn't hold the table tags/fields data in the abstract properties. I have added the property to store the table into this abstract class and updated `getValue` function to search for the correct fields/tags.

### DIFF
--- a/giraffe/src/components/geo/processing/NativeGeoTable.ts
+++ b/giraffe/src/components/geo/processing/NativeGeoTable.ts
@@ -8,10 +8,14 @@ import {
   GEO_HASH_COLUMN,
   LAT_COLUMN,
   LON_COLUMN,
+  START_COLUMN,
+  STOP_COLUMN,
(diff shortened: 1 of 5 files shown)


giraffe/src/components/geo/processing/NativeGeoTable.ts
```diff
@@ -8,10 +8,14 @@ import {
   GEO_HASH_COLUMN,
   LAT_COLUMN,
   LON_COLUMN,
+  START_COLUMN,
+  STOP_COLUMN,
   TABLE_COLUMN,
+  TIME_COLUMN,
```
hoorayimhelping, line 14: :+1: alphabetization

hoorayimhelping (approved):

zoesteinkamp (approved):
yay new tooltip

Issue #630, opened 2021-06-30
React hooks: best practices / fix all lint errors


TCL735:
Giraffe currently has 10 lint errors when checked by `eslint-plugin-react-hooks`

**EDIT**: this issue has been split into 2 separate issues. Only the lint errors remain in this issue. All lint warnings have been moved into a separate issue.

Two action items:
- [x] fix all lint errors (see list below)
- [x] activate the `eslint-plugin-react-hooks` rule by installing it

The lint errors are:

- [x] /giraffe/src/components/Geo.tsx
  46:3   error    React Hook "useEffect" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return?  react-hooks/rules-of-hooks
  51:53  error    React Hook "useState" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return?   react-hooks/rules-of-hooks
  63:3   error    React Hook "useEffect" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return?  react-hooks/rules-of-hooks

- [x] giraffe/src/components/MultiGrid.tsx
  601:28  error    React Hook "useForceUpdate" is called in function "forceUpdate" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter  react-hooks/rules-of-hooks

- [x] /giraffe/src/components/Plot.tsx
  16:19  error  React Hook "useRef" is called conditionally. React Hooks must be called in the exact same order in every component render  react-hooks/rules-of-hooks
  17:20  error  React Hook "useRef" is called conditionally. React Hooks must be called in the exact same order in every component render  react-hooks/rules-of-hooks

- [x] /giraffe/src/components/geo/CircleMarkerLayer.tsx
  83:9  error    React Hook "useLayoutEffect" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function  react-hooks/rules-of-hooks

- [x] /giraffe/src/components/geo/PointMapLayer.tsx
  71:9  error    React Hook "useLayoutEffect" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function  react-hooks/rules-of-hooks

- [x] /giraffe/src/utils/legend/useTooltipElement.ts
  33:5  error  React Hook "useAnnotationStyle" is called conditionally. React Hooks must be called in the exact same order in every component render  react-hooks/rules-of-hooks
[...]

PR #633, opened 2021-07-01
WIP: chore: disable eslint warning for missing dependency because it should not be included

TCL735:
As part of #630 

Disable the eslint warning proactively because we do not need `renderEffect` as a dependency. `renderEffect` is a function given by the consumer to be run inside the hook. The function should not change. It is used only to adjust the height of the static legend automatically. The hook should only run when anything associated with the height changes.

Note: this change is being made to eliminate a warning that would appear only if `eslint-plugin-react-hooks` is installed

### DIFF
--- a/giraffe/src/components/StaticLegendBox.tsx
+++ b/giraffe/src/components/StaticLegendBox.tsx
@@ -61,6 +61,7 @@ export const StaticLegendBox: FunctionComponent<StaticLegendBoxProps> = props =>
   ])
 
   const {style = {}} = staticLegendOverride
+  const {renderEffect} = staticLegendOverride
 
   const {
     legendBackgroundColor: backgroundColor,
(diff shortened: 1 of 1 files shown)


giraffe/src/components/StaticLegendBox.tsx
```diff
@@ -90,6 +91,10 @@ export const StaticLegendBox: FunctionComponent<StaticLegendBoxProps> = props =>
       headerTextMetrics,
       sampleTextMetrics,
     })
+    /* run this hook only for the height changes because the purpose is to set the height for user
+     * therefore, the dependencies should not include anything that doesn't affect height
+     */
+    // eslint-disable-next-line
```
hoorayimhelping, line 97: i think it's probably better from a code correctness and documentation standpoint to disable the specific rule, rather than the whole line.

TCL735: Yeah the problem is that without the plugin installed, disabling a specific rule that doesn't exist is in itself a lint error

TCL735: Since the original problem is a warning and not an error, maybe we should hold off on this until the plugin is actually installed

TCL735:
Closing this. Will redo undo new issue when time allows.

PR #635, opened 2021-07-06
test: geo utils

### LINKED ISSUE #579: Refactor: Maps Geo in Giraffe to fail gracefull
Currently, few calculations that are happening in Giraffe, do not throw an error and fail horribly in the UI. When such things happen, we want to send proper worded message to the UI to handle errors. For this refactor, we will need to write some tests and then find any error use cases and fail gracefully (for example: in getLatLon function).

Palakp41:
Closes #579 

Adding tests around all the maps/geo utils with different use cases

### DIFF
--- a/giraffe/src/utils/geo.test.ts
+++ b/giraffe/src/utils/geo.test.ts
@@ -1,6 +1,13 @@
 import {InfluxColors} from '../constants/colorSchemes'
-import {GeoQueryVariables, GeoViewLayer} from '../types/geo'
-import {calculateVariableAssignment, getMinZoom, getRowLimit} from './geo'
+import {GeoCircleViewLayer, GeoQueryVariables, GeoViewLayer} from '../types/geo'
+import {
+  calculateVariableAssignment,
+  formatCircleMarkerRowInfo,
+  getMinZoom,
+  getRowLimit,
+} from './geo'
+// import {geoTable} from '../../../stories/src/data/geoLayer'
+import {preprocessData} from '../components/geo/processing/tableProcessing'
 
 const pointMapLayer: GeoViewLayer[] = [
   {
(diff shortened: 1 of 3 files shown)


giraffe/src/utils/geo.test.ts
```diff
@@ -0,0 +1,126 @@
...
+  colorField: 'duration',
+  colors: [
+    {type: 'min', hex: '#ff00b3'},
+    {value: 50, hex: '#343aeb'},
+    {type: 'max', hex: '#343aeb'},
+  ],
+}
+
+describe('getMinZoom', () => {
+  it('can return minimum zoom for viewing maps', () => {
```
hoorayimhelping, line 44: are there more zoom mappings to width? or is 500 px wide => level 1 zoom the only one there is?

giraffe/src/utils/geo.test.ts
```diff
@@ -0,0 +1,126 @@
...
+    const width = 500
+    const minimumZoom = getMinZoom(width)
+    expect(minimumZoom).toBe(1)
+  })
+})
+
+describe('getRowLimit', () => {
+  it('will return the layer limit based on layer type', () => {
+    const layerLimit = getRowLimit([pointMapLayer])
```
hoorayimhelping, line 53: do only point maps have layer limits, or do all if them?

hoorayimhelping (approved):
looks good! a couple of questions about adding more tests

giraffe/src/utils/geo.test.ts
Palakp41: I am not sure, I don't see width being used anywhere in the ui or in the storybook. I have added the test so that if the formula to calculate minZoom changes, it will get caught here.

giraffe/src/utils/geo.test.ts
Palakp41: I will add more layers here

PR #636, opened 2021-07-09
fix: update native geo table to use s2 column

Palakp41:
Turns out there was a hardcoded value being used for using `s2_cell_id` to get lat/lon coordinates. I updated NativeGeoTable to use the selected s2Column

### DIFF
--- a/giraffe/package.json
+++ b/giraffe/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/giraffe",
-  "version": "2.15.8",
+  "version": "2.15.9",
   "main": "dist/index.js",
   "module": "src/index.js",
   "license": "MIT",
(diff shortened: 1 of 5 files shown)


giraffe/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/giraffe",
-  "version": "2.15.8",
+  "version": "2.15.9",
```
hoorayimhelping, line 3: i think we also need to update `giraffe/stories/package.json`'s version number.

@TCL735 is that true?

hoorayimhelping (approved):

giraffe/package.json
TCL735: Yes, ideally everyone should update it. But as long as one person (me) remembers to update it periodically, i think it's fine. I can't seem to find where `giraffe/stories/package.json`'s version number is actually displayed other than in that file, ie why it's useful.

Whereas the version number in `giraffe/giraffe/package.json` allows us to publish correctly.

TCL735 (approved):

giraffe/package.json
Palakp41: Updated the version number in `giraffe/stories/package.json`

PR #637, opened 2021-07-09
fix: fromFlux accounted for +/- Inf values

mtian725:
fixes: https://github.com/influxdata/ui/issues/804

#### Before
Screen Shot 2021-07-08 at 8 40 26 PM

#### After 
Screen Shot 2021-07-08 at 8 38 16 PM


TCL735:
See https://github.com/influxdata/giraffe/pull/505

giraffe/src/utils/fromFlux.ts
```diff
@@ -248,6 +248,10 @@ const parseValue = (value: string | undefined, columnType: ColumnType): any => {
     return undefined
   }
 
+  if (value === '+Inf' || value === '-Inf') {
+    return value
+  }
+
```
TCL735, line 254: What if in the future Infinity is expressed differently, for example: ∞

This will lead to the same problem again.

I made a suggestion for this fix in this pull request: https://github.com/influxdata/giraffe/pull/505

TCL735 (changes requested):

mtian725:
`Inf` values still work for all graph types. Except there are visual differences. Also, cleared up a case where visualization in the explore window crashed:

- Guage graph where the latest row of data was `+/-Inf` would cause the view window to crash. Fix causes the windows to displays a message instead and leaves it working.
#### Before (with fromFlux() turning `Inf` to `NaN`) 
image
#### After (with fromFlux() leaving `Inf` as  a string)
Screen Shot 2021-07-13 at 6 09 33 PM


## Line Graph for example
#### Before (with fromFlux() turning `Inf` to `NaN`) 
image

#### After (with fromFlux() leaving `Inf` as  a string)
Screen Shot 2021-07-13 at 5 54 52 PM
- Note that there are no hover points in the middle of the line. The two rows of data with `+Inf` were omitted and ignored when the line is formed.

I think this is fine due to data points of `Inf` being a result of bad data.

PR #639, opened 2021-07-13
feat: add default handlers to controlled mode

### LINKED ISSUE #461: Trigger zoom programmatically
I can't seem to find a way to trigger the zoom update if i provide a custom `onSetXDomain` callback in the config object.
I would like to be able to call something like `handleXBrushEnd`.
I need this for 2 reasons:

- First because the only way to set the `xDomain` property is to provide also the `onSetXDomain`, so i don't know how to zoom if i want to provide an initial `xDomain`.

- Second, would be handy to have the possibility to "sync" the domain of multiple plots, changing it for every one of them whenever a zoom event occurs on any.

Is this feasible?

EDIT

A possibility would be to add an `InteractionHandler` for the zoom, like in https://github.com/influxdata/giraffe/pull/417, and a `setDomain` function in the `InteractionHandlerArguments`.

### LINKED ISSUE #638: Documentation: domain controlled mode requirements
Current documentation for xDomain and yDomain mention a "controlled mode" which is not easily activated. Just passing an xDomain or a yDomain does not trigger controlled mode. There are other requirements which aren't mentioned.

See https://community.influxdata.com/t/giraffe-library-min-max-domains-and-configuration-of-layers/20717 as an example of an attempt at controlled mode without success but follows the current documentation.

This is also related to #461 - which is the behavior rather than the documentation. Also, #461 is the next logical step because after documentation is provided, the user would like a way to trigger the default zoom/un-zoom even with controlled mode.

TCL735:
Closes #461 
Closes #638 

Documents the necessary config options to use controlled mode. More importantly, adds default handlers to allow controlled mode to be more useful (ie, it now has access to zoom and un-zoom default behavior).

https://user-images.githubusercontent.com/10736577/125536164-4610943e-a7c4-4abd-8887-e5bf404e9695.mp4


giraffe/README.md
```diff
@@ -362,17 +362,25 @@ When using the comma separated values (CSV) from the Flux query as the `fluxResp
...
+- **xDomain**: _array[min, max]. Optional. When used, must also include at least 2 additional properties. 1) Must also include **onSetXDomainDefault** or **onSetXDomain**, or both. 2) Must also include **onResetXDomainDefault** or **onResetXDomain**, or both._ The x domain of the plot can be explicitly set with numbers denoting a minimum and a maximum value for the x-axis. If this option is passed, both _min_ and _max_ are required to be numbers, making the `<Plot>` operate in a "controlled" mode, where it always uses the passed x domain. Any brush interaction with the `<Plot>` that should change the x domain will use Giraffe's built in default handlers if `onSetXDomainDefault` is `true`. Additionally, the `onSetXDomain` callback function will be invoked if provided. Double clicking the plot will call Giraffe's built in default handers if `onResetXDomainDefault` is `true`. Additionally, the `onResetXDomain` callback function will be invoked if provided. If the `xDomain` option is not passed, then the component is "uncontrolled". It will compute, set, and reset the `xDomain` automatically.
```
hoorayimhelping, line 365: Is it possible to configure the default handlers by only passing in a function, rather than a function and a boolean? Is it possible to infer the intent from the value of the function either being a function or `undefined`?

I'm thinking of this from the perspective of having to modify the state correctly to have this work

hoorayimhelping (commented):
A suggestion on variable naming and a question about the need to use a boolean

giraffe/README.md
TCL735: Yes, it is possible just to pass in a function. Or just the boolean, or even both. They are independent of each other. The behavior here is additive.

giraffe/src/types/index.ts
```diff
@@ -55,21 +55,30 @@ export interface Config {
...
+  //     will use the default handler if `onSetXDomainDefault` is true,
+  //     and call `onSetXDomain` if provided.
+  //   - Double clicking the plot will use the default handler if `onResetXDomain`
+  //     is true, and call `onResetXDomain` if provided.
+  //
+  // When the `xDomain` option is not passed, then the
   // component is "uncontrolled". It will compute, set, and reset the `xDomain`
   // automatically.
   xDomain?: number[]
+  onSetXDomainDefault?: boolean
```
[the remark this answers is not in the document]
TCL735: Good call. Will update it.

giraffe/README.md
TCL735: We need both because there may be a use case where it is desirable to suppress the default handlers and just use the callback function. The callback function is not a real handler though, because it doesn't have access to private fields in `PlotEnv`

TCL735:
Update: i changed the config property for the default handlers to a single property, since the default handlers aren't really useful unless you use them all together. There really isn't a use case for using one but not the other, except for causing buggy behavior.

hoorayimhelping (approved):

TCL735:
The original test in PlotEnv.test is passing again without alteration. Bonus: a new test was added that tests new behavior which includes resetting the domains.

PR #641, opened 2021-07-16
fix: annotations much easier to click on

jrenee42:
easier clicking now that the hover legend is hidden when the mouse is on a clickable annotation top (either the range 'hat' or the point 'diamond')

if the legend is already hidden then this is a no-op.  nothing happens.  (because there is nothing to hide)

for ui#1119

### DIFF
--- a/giraffe/src/components/AnnotationLine.tsx
+++ b/giraffe/src/components/AnnotationLine.tsx
@@ -75,6 +75,8 @@ export const AnnotationLine: FunctionComponent<AnnotationLineProps> = props => {
   const clampedStart = Math.max(1, Math.round(startValue))
   const clampedEnd = Math.max(1, Math.round(stopValue))
 
+  console.log('ack ack cak 776a')
+
   if (dimension === 'y') {
     return (
       <>
@@ -116,14 +118,6 @@ export const AnnotationLine: FunctionComponent<AnnotationLineProps> = props => {
     )
   }
 
-  // dimension is x:
-  const handleMouseEnter = () => {
-    onHover(true)
-  }
-
-  const handleMouseLeave = () => {
-    onHover(false)
-  }
   /**
    * This is the rectangle or 'hat' that goes on top of a range annotation.
    * The entire hat is click to edit for the annotation.
@@ -136,9 +130,6 @@ export const AnnotationLine: FunctionComponent<AnnotationLineProps> = props => {
   const makeRangeRectangle = () => {
     const pixelMargin = 1
 
-    // cannot just directly reference the handleMouseEnter/Leave functions;
-    // that doesn't work (it gets called immediately).
-    // so need the arrow function when using createElement
     return createElement('polygon', {
       points: `${clampedStart - pixelMargin}, 0
           ${clampedEnd + pixelMargin}, 0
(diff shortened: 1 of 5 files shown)


hoorayimhelping:
@jrenee42 can you post a video or gif of this in action locally please?

jrenee42:
https://user-images.githubusercontent.com/3900960/126002287-91fedcce-52e1-474d-9fc0-6c0008f221d1.mov

showing on both xy/line graph and bandplot.  it's all about the annotation layer, so if more layers get annotations, this fix is agnostic and will keep working.

jrenee42:
> @jrenee42 can you post a video or gif of this in action locally please?

done.  see above

giraffe/src/components/AnnotationLine.tsx
```diff
@@ -128,6 +147,12 @@ export const AnnotationLine: FunctionComponent<AnnotationLineProps> = props => {
       fill: color,
       id: props.id,
       style: {cursor: 'pointer', opacity: '60%'},
+      onMouseEnter: () => {
+        handleMouseEnter()
+      },
+      onMouseLeave: () => {
+        handleMouseLeave()
+      },
```
TCL735, line 155: Since there is nothing unique about each `onMouseEnter` and `onMouseLeave` function, there is no need to create a separate copy of each function every time `makeRangeRectangle` is called. Ultimately, you just want to call `onHover(true)` or `onHover(false)`

I would remove the unnecessary wrapping
```
      onMouseEnter: handleMouseEnter,
      onMouseLeave: handleMouseLeave,
```

giraffe/src/components/AnnotationLine.tsx
```diff
@@ -120,6 +136,9 @@ export const AnnotationLine: FunctionComponent<AnnotationLineProps> = props => {
   const makeRangeRectangle = () => {
     const pixelMargin = 1
 
+    // cannot just directly reference the handleMouseEnter/Leave functions;
+    // that doesn't work (it gets called immediately).
+    // so need the arrow function when using createElement
```
TCL735, line 141: Only if you call `handleMouseEnter` immediately. If you don't call the function immediately, it doesn't get called immediately.

giraffe/src/components/AnnotationLine.tsx
```diff
@@ -166,6 +191,12 @@ export const AnnotationLine: FunctionComponent<AnnotationLineProps> = props => {
           style: {cursor: 'pointer'},
           id: props.id,
           className: 'giraffe-annotation-click-target',
+          onMouseEnter: () => {
+            handleMouseEnter()
+          },
+          onMouseLeave: () => {
+            handleMouseLeave()
+          },
```
TCL735, line 199: Since there is nothing unique about each `onMouseEnter` and `onMouseLeave` function, there is no need to create a separate copy of each function every time `makeRangeRectangle` is called. Ultimately, you just want to call `onHover(true)` or `onHover(false)`

I would remove the unnecessary wrapping
```
      onMouseEnter: handleMouseEnter,
      onMouseLeave: handleMouseLeave,
```

TCL735 (changes requested):

giraffe/src/components/AnnotationLine.tsx
jrenee42: well, when I just put only the function name it errored out.  using the fat arrows fixed it.

I tried to find documentation on createElement but could not find it.

TCL735: `onMouseEnter: handleMouseLeave,` works, same for Leave

were you referring to `handleMouseLeave,` ? there's no property called `handleMouseLeave` on the object passed to createElement

TCL735 (approved):

PR #642, opened 2021-07-19
chore: add CSS + types for SimpleTable

mtian725:
#632

Just CSS and types for PR #632. Nothing is using this at the moment.


giraffe/src/types/index.ts
```diff
@@ -453,6 +455,12 @@ export interface ScatterLayerConfig {
   symbol?: string[]
 }
 
+export interface SimpleTableLayerConfig {
+  type: 'simple table' // do not refactor or restrict to LayerTypes.Mosaic
```
hoorayimhelping, line 459:
```suggestion
  type: 'simple table' // do not refactor or restrict to LayerTypes.SimpleTable
```
@TCL735 what do these comments even mean?

giraffe/src/types/index.ts
```diff
@@ -916,3 +924,44 @@ export interface AnnotationMark {
   pin: AnnotationPinType
   id?: string
 }
+
+export enum ComponentColor {
```
hoorayimhelping, line 928: @TCL735 is this too much of Clockface leaking into giraffe?

hoorayimhelping (commented):
seems good to me, but I want to let Timmy take a look at the ClockFace constants we're adding

giraffe/src/types/index.ts
TCL735: No, but they don't belong here either. I will make another comment.

giraffe/src/types/index.ts
```diff
@@ -916,3 +924,44 @@ export interface AnnotationMark {
...
+  Submit = 'submit',
+}
+
+export enum BorderType {
+  None = 'none',
+  Horizontal = 'horizontal',
+  Vertical = 'vertical',
+  Both = 'both',
+  All = 'all',
+}
```
TCL735, line 967: If these are used only by SimpleTable then they should be moved to `giraffe/src/components/SimpleTable/Table/index.tsx` or if needed elsewhere in the `SimpleTable` hierarchy then one folder up is fine too. These don't belong in the global index file though.

TCL735 (changes requested):

giraffe/src/components/SimpleTable/Types.tsx
```diff
@@ -0,0 +1,41 @@
...
+  }
+  
+  export enum BorderType {
+    None = 'none',
+    Horizontal = 'horizontal',
+    Vertical = 'vertical',
+    Both = 'both',
+    All = 'all',
+  }
+
```
hoorayimhelping, line 41: since there isn't any JSX in this file, it shouldn't be a .tsx file; just a .ts file.

you'll want to rename it to:

`types.ts`

types should be lowercase and it should have a .ts extension.

hoorayimhelping (changes requested):
The content looks good, but there are some structural nitpicks about the case and extensions of your files and not having newlines at the end of them.

files need newlines at the end of them, and there are several in this PR that don't have them. See https://stackoverflow.com/a/729795/92446.

Here's how to automatically add them when using VSCode:
https://newbedev.com/visual-studio-code-insert-new-line-at-the-end-of-files

hoorayimhelping (approved):
nice one!

giraffe/src/types/index.ts
```diff
@@ -255,6 +255,7 @@ export enum LayerTypes {
   Mosaic = 'mosaic',
   Table = 'table',
   Geo = 'geo',
+  SimpleTable = 'simple table',
```
TCL735, line 258: Not a big deal but this should have waited until there was code that uses this definition.

giraffe/src/types/index.ts
```diff
@@ -271,6 +272,7 @@ export type LayerConfig =
   | MosaicLayerConfig
   | TableGraphLayerConfig
   | GeoLayerConfig
+  | SimpleTableLayerConfig
```
TCL735, line 275: Same as above. Not a big deal but this should have waited until there was code that uses this definition.

giraffe/src/types/index.ts
```diff
@@ -453,6 +455,12 @@ export interface ScatterLayerConfig {
   symbol?: string[]
 }
 
+export interface SimpleTableLayerConfig {
+  type: 'simple table' // do not refactor or restrict to LayerTypes.SimpleTable
+  showAll: boolean
+  fluxResponse?: string
+}
+
```
TCL735, line 463: Same as above. Not a big deal but this should have waited until there was code that uses this definition.

TCL735 (approved):

PR #643, opened 2021-07-19
chore: pagination components and utils for SimpleTable

mtian725:
#632

Pagination component for #632. Still not functional nor it being used.  Depends on #642.


giraffe/src/components/SimpleTable/PaginationNav/Button/Button.scss.d.ts
```diff
@@ -0,0 +1,27 @@
...
+  'cf-button-spinner': string;
+  'cf-button-square': string;
+  'cf-button-stretch': string;
+  'cf-button-success': string;
+  'cf-button-tertiary': string;
+  'cf-button-warning': string;
+  'cf-button-xs': string;
+}
+export const cssExports: CssExports;
+export default cssExports;
```
TCL735, line 27: This file should not be added. It is auto-generated with `yarn build` when there is a consuming component.

TCL735 (dismissed):

giraffe/src/components/SimpleTable/PaginationNav/Button/Button.scss.d.ts
mtian725: `yarn build` throwing errors saying that `./Button.scss` could not be found. However, with all the pieces together `yarn build` does work, so just going to remove this for now.

giraffe/src/components/SimpleTable/PaginationNav/Button/Button.scss
```diff
@@ -0,0 +1,423 @@
+
```
TCL735, line 1: Button.scss is HUGE. There's too many definitions and styles in there. Let's cut it down to only the ones we are using.

TCL735 (changes requested):

giraffe/src/components/SimpleTable/PaginationNav/Button/Button.scss
mtian725: Removed unnecessary CSS and cut it down by about half. Color and Size was removed because SimpleTable components are going to be static. Thus, it's not going to change beyond medium size and tertiary color.

mtian725: Loading and Disabled was removed for 2 reasons. 

1. SimpleTable does not pass in any props from SimpleTableGraph.tsx that would include ComponentStatus.Disabled or ComponentStatus.Loading

2. If for some reason it does use these statuses and it was missed, observations on the UI side shows that ComponentStatus.Disabled is often used when a faulty option or invalid input is given. ComponentStatus.Loading is often used after the action of pressing the button sends a request somewhere, and that wait time to process said request means that it is “loading”. Both situations should not happen for our uses of Button.

mtian725: Note: `yarn built` ran but did not update the `d.ts` file.

TCL735 (approved):

PR #645, opened 2021-07-21
fix: delete manually copied files

TCL735:
Part of #632 

These files should be deleted, because they were manually copied in https://github.com/influxdata/giraffe/pull/642 - but there is no component that consumes the corresponding `.scss` file, therefore the build step should not have generated these 3 files.

Once a component is added, then run `yarn build` and a `.scss.d.ts` will be created for any consumed `.scss` file. At that time, add the newly auto-generated file to be committed. But not before. **By following this procedure, we cut down on unknown and unused `.scss.d.ts`**.

--

As a test, do these steps to confirm that all consumed `.scss` always auto-generate:
- delete any long-standing `.scss.d.ts` that has a consumer such as `/giraffe/src/components/AnnotationLine.scss.d.ts`
- confirm visually that the file is deleted with `git status`
- at the command prompt, in the `/giraffe` folder, run `yarn build`
- you should see that the file you deleted in the first step has been restored

But if you run the above steps with any of the `.scss.d.ts` files in the SimpleTable folder, they do not get re-generated, because there is no consumer.


mtian725 (approved):

hoorayimhelping (approved):
thanks for the explanation on this

PR #649, opened 2021-07-29
fix: formatting numbers in the tableGraph

jrenee42:
reusing the formatter from the single stat layer

  * added test
  * changed data for the storybook to have larger numbers to show commas


hoorayimhelping:
is this intended to be released immediately or can it sit in master? do we need to bump the version?

jrenee42:
> is this intended to be released immediately or can it sit in master? do we need to bump the version?

this doesn't need immediate release.

stories/src/data/tableGraph.ts
```diff
@@ -4,8 +4,8 @@ export const tableCSV = `#group,false,false,true,true,false,false,true,true,true
 ,result,table,_start,_stop,_time,_value,_field,_measurement,cpu,host
 ,,0,2020-07-01T21:45:43.0968Z,2020-07-01T21:50:43.0968Z,2020-07-01T21:45:50Z,2.2,usage_system,cpu,cpu1,MBP15-TLUONG.local
 ,,0,2020-07-01T21:45:43.0968Z,2020-07-01T21:50:43.0968Z,2020-07-01T21:46:00Z,1,usage_system,cpu,cpu1,MBP15-TLUONG.local
-,,0,2020-07-01T21:45:43.0968Z,2020-07-01T21:50:43.0968Z,2020-07-01T21:46:10Z,1.2,usage_system,cpu,cpu1,MBP15-TLUONG.local
-,,0,2020-07-01T21:45:43.0968Z,2020-07-01T21:50:43.0968Z,2020-07-01T21:46:20Z,0.8,usage_system,cpu,cpu1,MBP15-TLUONG.local
+,,0,2020-07-01T21:45:43.0968Z,2020-07-01T21:50:43.0968Z,2020-07-01T21:46:10Z,12123,usage_system,cpu,cpu1,MBP15-TLUONG.local
+,,0,2020-07-01T21:45:43.0968Z,2020-07-01T21:50:43.0968Z,2020-07-01T21:46:20Z,2345678,usage_system,cpu,cpu1,MBP15-TLUONG.local
```
TCL735, line 8: Thanks for adding the two lines that will show the commas.

I don't think we should replace any of the existing data, though. The existing data will help with identifying any issues with formatting decimal places (regression). The table will work just fine with more lines of data.

giraffe/src/components/TableCell.tsx
```diff
@@ -167,16 +167,16 @@ const getContents = (props: Props): string => {
...
     isFieldName(isVerticalTimeAxis, rowIndex, columnIndex)
   ) {
     return defaultTo(getFieldName(props), '').toString()
   }
 
-  if (!isNaN(+data) && decimalPlaces.isEnforced && decimalPlaces.digits < 100) {
-    return (+data).toFixed(decimalPlaces.digits)
+  if (!isNaN(+data)) {
+    // method needs the first arg to be a number to work properly
+    return formatStatValue(+data, {decimalPlaces})
```
TCL735, line 179: The whole function `getContents` probably needs unit tests to ensure that when `data` is `null` or empty string, the desired output is given.

Coercion is happening here to accommodate both `1234` and `"1234"` (as examples), but then coercion fails when we have an empty string.

TCL735 (changes requested):

giraffe/src/components/TableCell.tsx
TCL735: And to clarify: an empty string should show as an empty string.

Right now, and even with this code change, it will show as `0` which is not accurate. (This is the failure that I referred to above.)

jrenee42: don't need to fix the !Nan coercion.  but yes, empty string should yield empty string from this function.  adding tests and fixing that problem *before* it gets to this clause.

giraffe/src/components/TableCell.tsx
```diff
@@ -164,19 +169,23 @@ const getContents = (props: Props): string => {
   } = props
   const {decimalPlaces} = properties
 
+  if (!data || (isString(data) && isBlank(data))) {
+    return ''
```
TCL735, line 173: `return String(data)`

Let's just give back what the value was that caused this edge case.

giraffe/src/components/TableCell.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+    testNumber('.8', '0.80')
+    testNumber('1.8', '1.80')
+    testNumber('.822', '0.82')
+    testNumber('2244', '2,244.00')
+  })
+  it('should just pass through an empty string', () => {
+    const tableCellProps = {data: '', dataType: 'string', ...baseProps}
+
+    const result = getContents(tableCellProps)
+    expect(result).toEqual('')
```
TCL735, line 54: `expect(result).toEqual(tableCellProps.data)`

Due to the suggested change in TableCell, it's better to assert the given value.

giraffe/src/components/TableCell.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+    const tableCellProps2 = {data: '  ', dataType: 'string', ...baseProps}
+
+    const result2 = getContents(tableCellProps2)
+    expect(result2).toEqual('')
```
TCL735, line 59: `expect(result2).toEqual(tableCellProps2.data)`

Due to the suggested change in TableCell, it's better to assert the given value.

TCL735 (changes requested):

TCL735 (approved):

PR #652, opened 2021-08-06
fix: remove conditional usage of useRef hook

TCL735:
Part of #630 

Resolves the lint error for

```
/giraffe/src/components/Plot.tsx
16:19 error React Hook "useRef" is called conditionally. React Hooks must be called in the exact same order in every component render react-hooks/rules-of-hooks
17:20 error React Hook "useRef" is called conditionally. React Hooks must be called in the exact same order in every component render react-hooks/rules-of-hooks
```

### DIFF
--- a/giraffe/src/components/Plot.tsx
+++ b/giraffe/src/components/Plot.tsx
@@ -10,12 +10,20 @@ interface PlotProps {
   layerCanvasRef?: RefObject<HTMLCanvasElement>
 }
 
-export const Plot: FunctionComponent<PlotProps> = ({
-  children,
-  config,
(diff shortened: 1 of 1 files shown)


TCL735:
Sanity check looks good. All graphs behaving as expected.

TCL735:
...and screenshot utility still works too.

hoorayimhelping (approved):
it really annoys me that hooks force us to change code that works for our style and our domain model into code that works for react's internal model of rendering. in this case, we have to write more code to do what the language already gives us (and so is more likely to be buggy) to use hooks, which provide the same functionality as classes. we're effectively doubling our effort and increasing bugs in the codebase for what amounts to small, stylistic benefits.

PR #653, opened 2021-08-06
fix: remove unnecessary wrapped function to resolve lint error

TCL735:
Part of #630 

Resolves the lint error for
```/giraffe/src/components/MultiGrid.tsx
601:28 error React Hook "useForceUpdate" is called in function "forceUpdate" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter react-hooks/rules-of-hooks```


TCL735:
Sanity check looks good. Table graph still working as expecting.

Palakp41 (approved):

PR #654, opened 2021-08-07
fix: move useLayoutEffect into component to resolve lint error

TCL735:
Part of #630 

Resolves lint errors for
``` 
/giraffe/src/components/geo/CircleMarkerLayer.tsx
83:9 warning React Hook useLayoutEffect received a function whose dependencies are unknown. Pass an inline function instead react-hooks/exhaustive-deps
83:9 error React Hook "useLayoutEffect" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function react-hooks/rules-of-hooks

 /giraffe/src/components/geo/PointMapLayer.tsx
71:9 warning React Hook useLayoutEffect received a function whose dependencies are unknown. Pass an inline function instead react-hooks/exhaustive-deps
71:9 error React Hook "useLayoutEffect" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function react-hooks/rules-of-hooks
```

### DIFF
--- a/giraffe/src/components/geo/CircleMarkerLayer.tsx
+++ b/giraffe/src/components/geo/CircleMarkerLayer.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {createRef, FunctionComponent, useLayoutEffect} from 'react'
+import React, {createRef, FunctionComponent} from 'react'
 import {CircleMarker} from 'react-leaflet'
 
 // Utils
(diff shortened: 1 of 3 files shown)


TCL735:
This one will need some testing from the experts on the Maps feature.

Palakp41 (approved):
Looks good, thanks for fixing these errors :)

PR #655, opened 2021-08-07
fix: move early return after all hooks

TCL735:
Part of #630 

Resolves the lint error for
```
/giraffe/src/components/Geo.tsx
46:3 error React Hook "useEffect" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return? react-hooks/rules-of-hooks
51:53 error React Hook "useState" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return? react-hooks/rules-of-hooks
63:3 error React Hook "useEffect" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return? react-hooks/rules-of-hooks
```

### DIFF
--- a/giraffe/src/components/Geo.tsx
+++ b/giraffe/src/components/Geo.tsx
@@ -39,11 +39,11 @@ const Geo: FunctionComponent<Props> = props => {
   } = props
   const {layers, tileServerConfiguration} = props
   const {tileServerUrl, bingKey} = tileServerConfiguration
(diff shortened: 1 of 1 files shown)


TCL735:
This one requires some testing from experts on the Maps feature.

Palakp41 (approved):

Issue #661, opened 2021-08-21
Update icomoon fonts


ChitlangeSahas:
Our icomoon files are too old and stale and we need to update them to match them up with clockface.

PR #662, opened 2021-08-21
chore: update giraffe fonts

ChitlangeSahas:
Closes: #661 

We need to update our font files so that they are up to date with the UI and clockface's. Here is a detailed reasoning to why: 

@mtian725 worked on porting the Simple table to the Giraffe and then consume that in the UI. Eveerything was fine except there was an unexpected behavior with the icons showing up differently depending on where the `SimpleTable` component was being consumed. 

In storybook, it was fine, the left and right caret showed up as expected. But when we consumed the component in Giraffe, they turned into these: 

screen_shot_2021-08-12_at_7 02 02_pm

One noticeable thing here was they were not random, the icons were consistently the wrong ones instead of the left and the right carets. So this had to be coming from somewhere and it was not magically happening. 

I pretty much dissected every file related to icons in Giraffe and eventually found that the font files were outdated and this was causing a conflict when consumed in the UI. 

As seen here, font files in clockface were last updated 25 days ago and in the giraffe it has been almost a year. 

Clockface:

Screen Shot 2021-08-21 at 10 54 21 AM

Giraffe:

Screen Shot 2021-08-21 at 10 54 50 AM

Now works fine in the UI: 

Screen Shot 2021-08-21 at 10 59 33 AM


giraffe/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/giraffe",
-  "version": "2.18.0",
+  "version": "2.18.1",
```
ChitlangeSahas, line 3: Bumped version for publishing

giraffe/src/components/Icon.scss
```diff
@@ -29,6 +29,58 @@
...
+    content: '\e902';
+  }
+  &.shield-solid:before {
+    content: '\e903';
+  }
+  &.lock:before {
+    content: '\e900';
+  }
+  &.heart-solid:before {
+    content: '\e96f';
```
ChitlangeSahas, line 49: This entire file came from Clockface, [[L20](https://github.com/influxdata/clockface/blob/master/src/Styles/icon.scss#L20)]

TCL735 (approved):

hoorayimhelping (approved):
works for me, but Timmy has the most context and knowledge about pulling files over from clockface, so make sure he looks over this

PR #665, opened 2021-08-24
fix: remove margins from sizedTable

ChitlangeSahas:


TCL735:
This is brilliant investigative debugging. As long as `RawFluxDataTable` and `Table` has no layout issues due to the missing margins, this is a great fix.

PR #673, opened 2021-09-08
fix: strings that are numbers still render as strings

ChitlangeSahas:
Closes: https://github.com/influxdata/idpe/issues/10645

Our table was formatting `string` dataType that were numbers, as numbers. This is because we use `isNaN` to make the decision, I added a layer around those checks to check whether the dataType of the flux column is `string` in the first place. 

Screen shots: 

`_value` column dataType `string`

Screen Shot 2021-09-08 at 10 40 45 AM

`_value` column dataType `double`
 
Screen Shot 2021-09-08 at 10 41 11 AM

### DIFF
--- a/giraffe/package.json
+++ b/giraffe/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/giraffe",
-  "version": "2.18.1",
+  "version": "2.18.2",
   "main": "dist/index.js",
   "module": "src/index.js",
   "license": "MIT",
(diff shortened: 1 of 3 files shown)


TCL735 (approved):

TCL735:
@ChitlangeSahas also - do we need a version bump?

hoorayimhelping (approved):

ChitlangeSahas:
> @ChitlangeSahas also - do we need a version bump?

Yes! I'll push one.

PR #678, opened 2021-09-17
fix: remove eot, ttf and svg font files for icomoon

ChitlangeSahas:
Attempt to fix: https://github.com/influxdata/ui/issues/2629

Those font files are not relevant for the web. This might be a cause of the `OTS parsing error` and the `font rejected by sanitizer` errors. Locally I was able to get rid of those by removing the incompatible/unnecessary files.


TCL735:
Closing this. Addressed in https://github.com/influxdata/giraffe/pull/684

PR #685, opened 2021-10-13
chore: remove duplicate rule for css in webpack config

TCL735:
Clean up. Sanity check looks good. Maps are loading in storybook with styles.

Screen Shot 2021-10-12 at 5 15 09 PM


TCL735:
> > bump package.json?
> 
> I could. It seemed too minor to kick out another version.

Done.

PR #699, opened 2021-11-13
chore: move Single Stat stories and add custom csv option

TCL735:
In preparation for possible future improvements to Single Stat, move the stories to their own file. Also, add a story that uses custom .csv

Screen Shot 2021-11-12 at 4 31 41 PM


hoorayimhelping (approved):

TCL735:
Force update to rebase with latest security updates. Using this PR as the opportunity to update the version number in preparation for publishing.

PR #709, opened 2021-11-16
chore: bring back config.js as preview.js to prevent vertical scrollbar

TCL735:
We do need this file but it should be named `preview.js`. It prevents the unsightly vertical scrollbar in storybook

### DIFF
--- a/stories/.storybook/preview.js
+++ b/stories/.storybook/preview.js
@@ -1,0 +1,5 @@
+import {addParameters} from '@storybook/react'
+
+addParameters({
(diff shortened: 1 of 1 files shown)


hoorayimhelping (approved):

hoorayimhelping:
my bad, I couldn't find the scrollbar when I looked. What stories does it show up on?

TCL735:
> my bad, I couldn't find the scrollbar when I looked. What stories does it show up on?

All of them, actually :grin:

Issue #697, opened 2021-11-01
Color map: When drawing data series, assign color using color preference object


kristinarobinson:
Using group key,  look up key in color preference map.

If present, then use index to look up color in color family.

If NOT present, use current index to assign new color in map, and look up color in color family.
  then increment current index.
  if current index == color family max
  then set current index = 0
  Trigger a save back to IDPE for color preference object.

PR #712, opened 2021-11-18
feat: use color mapping for Line layer

ChitlangeSahas:
closes: #697

Use the color map, when provided from the UI, for the Line Layer.

Proof that it works, override color map to be white, from UI:

* UI code overriding the map for testing.

Capture d’écran, le 2021-11-23 à 11 50 48 AM

* Giraffe Line layer and the hover layer respects the color map. 
 
Capture d’écran, le 2021-11-23 à 11 51 30 AM

 
* With multiple lines:

Capture d’écran, le 2021-11-24 à 11 51 24 AM
 
Also notice the points on the hover line cross hair, they respect the color mapping too. 

Capture d’écran, le 2021-11-24 à 11 52 33 AM

Also adds a storybook example.

Capture d’écran, le 2021-11-24 à 1 36 46 PM


giraffe/src/components/LineLayer.tsx
```diff
@@ -43,6 +43,10 @@ export const LineLayer: FunctionComponent<Props> = props => {
     shadeAboveY: height,
   }
 
+  if (config.colorMapping) {
+    config.colorMappingCallback(config.colorMapping)
+  }
+
```
ChitlangeSahas, line 49: We could delete this if this isn't how we want to use the callback.

giraffe/src/utils/legend/tooltip.ts
```diff
@@ -95,7 +96,11 @@ export const getPointsTooltipData = (
   const colors = orderDataByValue(
     hoveredRowIndices,
     sortOrder,
-    hoveredRowIndices.map(i => fillScale(groupColData[i]))
+    hoveredRowIndices.map(i =>
+      colorMapping
+        ? colorMapping.mappings[groupColData[i]].color
+        : fillScale(groupColData[i])
+    )
```
ChitlangeSahas, line 103: This is the tooltip part of logic for color mapping.

giraffe/src/components/LineLayer.tsx
```diff
@@ -43,6 +43,10 @@ export const LineLayer: FunctionComponent<Props> = props => {
     shadeAboveY: height,
   }
 
+  if (config.colorMapping) {
```
hoorayimhelping, line 46:
```suggestion
  if (config.colorMapping && config.colorMappingCallback) {
```

let's not infer that the `colorMappingCallback` was passed in if the `colorMapping` is, let's be explicit that they're both there

hoorayimhelping (commented):
nicely done!

I had one request about inferring the existence of a property.

I think this should be a minor version bump in giraffe/giraffe's and giraffe/stories' package.json. In other words, the new version should be `2.21.0`

Let's also let Timmy weigh in before you ship this.

giraffe/src/components/LineLayer.tsx
ChitlangeSahas: Very true! I'll fix that.

TCL735 (changes requested):
There should be a storybook that demonstrates this. Extending some existing stories to include an option for the color mapping would be enough.

TCL735:
Also, i think this won't work yet for Static Legend.

ChitlangeSahas:
> Also, i think this won't work yet for Static Legend.

@TCL735  correct, I would have to look at static legend's logic just like the tooltip, I assume.

stories/src/data/fluxCSV.ts
```diff
@@ -86,3 +86,901 @@ export const tooltipFalsyValues = `#group,false,false,true,true,true,false,false
...
+#default,mean,,,,,,,,
+,result,table,_start,_stop,_time,_value,_field,_measurement,sensor_id
+,,0,2021-11-24T18:43:49.742147732Z,2021-11-24T19:43:49.742147732Z,2021-11-24T18:44:00Z,0.3786237915662678,co,airSensors,TLM0100
+,,0,2021-11-24T18:43:49.742147732Z,2021-11-24T19:43:49.742147732Z,2021-11-24T18:44:10Z,0.37735360421915626,co,airSensors,TLM0100
+,,0,2021-11-24T18:43:49.742147732Z,2021-11-24T19:43:49.742147732Z,2021-11-24T18:44:20Z,0.38950567433178224,co,airSensors,TLM0100
+,,0,2021-11-24T18:43:49.742147732Z,2021-11-24T19:43:49.742147732Z,2021-11-24T18:44:30Z,0.3906371504590928,co,airSensors,TLM0100
+,,0,2021-11-24T18:43:49.742147732Z,2021-11-24T19:43:49.742147732Z,2021-11-24T18:44:40Z,0.3958325900427829,co,airSensors,TLM0100
+,,0,2021-11-24T18:43:49.742147732Z,2021-11-24T19:43:49.742147732Z,2021-11-24T18:44:50Z,0.41191279552390064,co,airSensors,TLM0100
+,,0,2021-11-24T18:43:49.742147732Z,2021-11-24T19:43:49.742147732Z,2021-11-24T18:45:00Z,0.4180948178942983,co,airSensors,TLM0100
+,,0,2021-11-24T18:43:49.742147732Z,2021-11-24T19:43:49.742147732Z,2021-11-24T18:45:10Z,0.42034671367558174,co,airSensors,TLM0100
```
ChitlangeSahas, line 101: A significant diff is from CSV data.

TCL735 (approved):

hoorayimhelping (approved):
Nice work!

PR #716, opened 2021-12-20
fix: Broken Tooltip

### LINKED ISSUE #680: Tooltip position affected by scroll
The tooltip position offset is somehow affected by the scroll.

Expectation is that tooltip all the time must be in correct position. 

https://user-images.githubusercontent.com/17004593/134341169-f4973718-698a-463e-bc9b-3c0d27033d24.mp4

Thank you

antondalgren:
When a page contained more graphs than the page could fit, the tooltip would disappear if scrolling down/to the right. This fix makes the Tooltip show for all graphs on these type of pages.

Closes #680

### DIFF
--- a/giraffe/src/utils/useMousePos.ts
+++ b/giraffe/src/utils/useMousePos.ts
@@ -14,7 +14,7 @@ interface UseMousePosProps {
 const mousePositionFromEvent = (e: MouseEvent<HTMLDivElement>) => {
   const {top, left} = e.currentTarget.getBoundingClientRect()
 
-  return {x: e.pageX - left, y: e.pageY - top}
+  return {x: e.clientX - left, y: e.clientY - top}
 }
 
 export const useMousePos = (): [MousePosition, UseMousePosProps] => {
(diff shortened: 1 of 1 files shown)


hoorayimhelping:
@antondalgren Thank you for looking into this. The code seems reasonable to me; I still need to do some testing within the application. And I'd like to give @TCL735 a chance to look it over as well, but he's going to be out for a few days.

antondalgren:
Thank you for the swift response. Yeah, I'm not sure if anything else that was relying on this hook has been affected negatively. But for the tooltip, https://github.com/antondalgren/giraffe/commit/b3ce5f064d29b9b1173aecec046b489501b8bd95#diff-b482bfba8e33e91e667830e369d09a33eba5c6e7ce2b459715f910bfe561005eL53 this line was pointing to the `<body>` tag, causing the calculation to become wrong once the start of the body tag was being outside of the screen.

iamrekas:
Hey, thank you for make this PR to fix the issue #680. Maybe somebody can review this PR and merge?

TCL735:
This is good. I will approve and merge.

We will also publish a new version with this change. As this pull request doesn't have the version number bumps, the new version will be created from a follow-on pull request.

TCL735 (approved):

PR #723, opened 2022-01-13
fix: use Intl.NumberFormat to properly format in the user's locale

TCL735:
Part of https://github.com/influxdata/ui/issues/3519

- Existing logic for number formatting is trying too hard and failing to account for all locales.
- Use `Intl.NumberFormat` to do all the heavy lifting.
- Dead on accurate due the presence of unit tests in `giraffe/src/utils/formatStatValue.test.ts`


hoorayimhelping (approved):
nice one!

PR #728, opened 2022-01-22
chore: upgrade Storybook but keep addon-knobs

TCL735:
Upgrade Storybook:

- to resolve dependabot alerts, which are entirely for Storybook related dependencies
- keep `@storybook/addon-knobs` for now even though it is being deprecated in favor of `@storybook/addon-controls`
   - migrating to addon-controls involves changes to our stories which requires a bigger effort
- upgrade `typescript` because `react/router`, which is used by the latest Storybook, needs it otherwise a Typescript error will fail all builds

### DIFF
--- a/giraffe/package.json
+++ b/giraffe/package.json
@@ -104,7 +104,7 @@
     "sass-loader": "^10.0.3",
     "style-loader": "^2.0.0",
     "ts-jest": "^26.5.0",
-    "typescript": "3.8.3",
+    "typescript": "4.5.5",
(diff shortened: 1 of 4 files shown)


hoorayimhelping (approved):
thanks for doing this, seems like a real pita

TCL735:
There is more to do here. eslint is failing and needs to be upgraded but even after upgrading there is some configuration issue.

TCL735:
This opened a huge can of worms. Too complex to do as a chore. Closing this.

PR #730, opened 2022-02-03
feat: enhance fromFlux utility to include all result column names

### LINKED ISSUE #726: fromFlux should name all of the result columns
Related to https://github.com/influxdata/ui/issues/3404

According to Flux documentation, [annotated CSV](https://docs.influxdata.com/influxdb/cloud/reference/syntax/annotated-csv), requires all of the header rows and annotation rows. A header row example looks like this: **result, table, _start, _stop, _time, _value, _field, _measurement**

The **result** column in the data is always blank. Instead, **result** relies on the `#default` annotation row to determine the column name, often defaulted as "_result". In some queries, **result** may be something else, examples include "mean", "median", "max", "min", "count", and so on. Basically, any name that may come from the `yield` function in Flux. 

Certain visualizations, particularly Band, look for options-specified column names in order to render the graph. Band relies on the user to supply the correct column names, and Giraffe makes the assumption that the user interface would allow the user to correctly identify the column name based on the user's knowledge of the data being analyzed.

However, this is not very user-friendly and requires a certain sophistication, both on the user's part and especially on the implementation details of the user interface.

### Proposal:
To assist with identifying the available **result** columns in annotated CSV, the `fromFlux` utility should also return all of the **result** column names, even when there is only a default column name.

TCL735:
Closes #726 

Adds a property to hold all of the result column names in the annotated CSV from the results of a Flux query. This allows the `fromFlux` utility to assist the developer in identifying valid result column names, which in turn can simplify user interface validation of visualization configuration options and properties.

### DIFF
--- a/giraffe/src/utils/fromFlux.ts
+++ b/giraffe/src/utils/fromFlux.ts
@@ -3,13 +3,18 @@ import {Table, ColumnType, FluxDataType} from '../types'
 import {assert} from './assert'
 import {newTable} from './newTable'
 
+const RESULT_COLUMN_INDEX = 1
 export interface FromFluxResult {
   error?: Error
+
   // The single parsed `Table`
   table: Table
 
   // The union of unique group keys from all input Flux tables
   fluxGroupKeyUnion: string[]
+
+  // All possible values of the "result" column
+  resultColumnNames: string[]
 }
 
 type Column =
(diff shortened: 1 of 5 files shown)


giraffe/src/utils/fromFlux.ts
```diff
@@ -158,14 +178,16 @@ export const fromFlux = (fluxCSV: string): FromFluxResult => {
     const result = {
       table,
       fluxGroupKeyUnion: Array.from(fluxGroupKeyUnion),
+      resultColumnNames: Array.from(resultColumnNames),
     }
 
     return result
   } catch (error) {
     return {
+      error: error as Error,
```
TCL735, line 187: We have to use `as Error` here because the `catch` clause must take **_any_** as the type.

hoorayimhelping (approved):

PR #734, opened 2022-02-09
fix: ensure Band's mainColumnName cannot be reused for lower and upper

TCL735:
Part of https://github.com/influxdata/ui/issues/3404

- Creates a normalizer to set the `config` including the `layers` to acceptable defaults and expectations. This is also the starting point for a separate bigger issue (see https://github.com/influxdata/giraffe/issues/447)

BEFORE, without normalization:

https://user-images.githubusercontent.com/10736577/153316844-83c3fd57-a520-4248-b77b-1eb6a42ca0de.mp4

AFTER, with normalization:

https://user-images.githubusercontent.com/10736577/153316876-29b6e61d-7cd0-4b2d-89d8-dc24a5ddc02d.mp4

### DIFF
--- a/giraffe/src/components/PlotResizer.tsx
+++ b/giraffe/src/components/PlotResizer.tsx
@@ -4,6 +4,7 @@ import {LayerTypes, PlotDimensions, SizedConfig} from '../types'
 
 import {get} from '../utils/get'
 import {resizePlotWithStaticLegend} from '../utils/legend/resizePlot'
+import {normalizeConfig} from '../utils/normalizeConfig'
 import {usePlotEnv} from '../utils/usePlotEnv'
 
 import {SizedPlot} from './SizedPlot'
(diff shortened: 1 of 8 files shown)


giraffe/src/utils/normalizeConfig.ts
```diff
@@ -0,0 +1,22 @@
...
+      if (!upperColumnName || upperColumnName === layerConfig.mainColumnName) {
+        upperColumnName = ''
+      }
+      if (!lowerColumnName || lowerColumnName === layerConfig.upperColumnName) {
+        lowerColumnName = ''
+      }
+      return {...layerConfig, upperColumnName, lowerColumnName}
+    }
+    return layerConfig
+  }) || []
```
hoorayimhelping, line 17: if this got passed in as a default argument, it might make the code a little simpler, since we wouldn't need to check if `layers` exists before calling map on it

giraffe/src/components/PlotResizer.tsx
```diff
@@ -28,16 +29,16 @@ export const PlotResizer: FC<PlotResizerProps> = props => {
...
-  const sizedConfig: SizedConfig = useMemo(
+  const normalizedConfig: SizedConfig = useMemo(
     () => ({
-      ...config,
+      ...normalizeConfig(config),
       height: resized.height,
       width: resized.width,
     }),
-    [config, height, width]
+    [config, height, width] // eslint-disable-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
TCL735: No, we can't put `normalizedConfig` into the dependency array, because this hook is the function that initializes `normalizedConfig`.

What's happening here is this: "if the user's sends in an updated configuration, or if the height and/or width the Plot changes, calculate a new normalized config"

giraffe/src/utils/normalizeConfig.ts
TCL735: Actually, I got a surprise. Making this change and then passing in `null` doesn't work.

I added a unit test to prove this. So, keep it as is, but with the requisite unit test to ensure

PR #328, opened 2020-10-21
fix: formatters allow for no formatting

zoesteinkamp:
This is needed to fix the differences in formatting, by allowing no formatting

PR #747, opened 2022-03-23
fix: allow siPrefixFormatter to ignore SI prefixes

TCL735:
Part of https://github.com/influxdata/ui/issues/3775

Previously, an [attempted fix](https://github.com/influxdata/giraffe/pull/328) put in the wrong logic for ignoring SI Prefixes (which groups large numbers into units like "K" for thousands and "M" for millions). This fixes that error and allows consuming applications to disable the SI Prefixes.

https://user-images.githubusercontent.com/10736577/159813185-70712a06-379d-4f80-bade-3546b74dead7.mp4


TCL735:
This works with InfluxData's Cloud UI:


https://user-images.githubusercontent.com/10736577/159816559-05ab8e89-1678-4580-bc2b-bd42af95514c.mp4

hoorayimhelping (approved):

PR #752, opened 2022-05-14
feat: improve fromFlux performance

### LINKED ISSUE #750: Spike: improve fromFlux function to be more performant
At its current iteration, the fromFlux function can be slow and bloated, causing any visualizations to lag behind CSV parsing. We should improve our parser to reduce the amount of memory its creating when parsing through CSVs AND try and reduce iterations within the CSVs in order to improve performance

asalem1:
Closes #750 

This PR brings a significant improvement to the `fromFlux` parser by doing two things:

1. updating the `splitChunks` function to use a more linear, less memory-intensive approach
2. removing the `parseValue` function for inline conditionals to reduce memory-usage of a function that's generated every time a deeply nested loop runs.

Before:
Screen Shot 2022-05-14 at 5 35 02 AM

After:
Screen Shot 2022-05-14 at 5 36 56 AM

### Reviewer Notes
While this solution does replace a like-for-like version of the fromFlux parser functionality, it should by no means be considered a final solution to the problem. The current implementation can and should be iterated upon. More specifically, the parsing process is about chunking the data, parsing it, restructuring it, and rejoining it. We should simplify this by using pointer references and getting the relevant data that way, rather than simply splitting, trimming, and rejoining strings in order to parse out relevant data (see the second screenshot where `trim` is takes an entire second)

### Follow-up items:
Since the `fromFlux` function is used both in the UI and in Giraffe, we should follow this work up by feature flagging the usage of `fromFlux` vs `fastFromFlux` in order to ensure that the parser works as expected and there are no breaking changes with this. This should be done by first feature flagging the usage within the UI. Once we know have successfully proven that no breaking changes have been introduced, we should roll out the `fastFromFlux` to be used inside Giraffe

### DIFF
--- a/giraffe/src/utils/fromFlux.ts
+++ b/giraffe/src/utils/fromFlux.ts
@@ -196,7 +196,46 @@ export const fastFromFlux = (fluxCSV: string): FromFluxResult => {
   let tableLength = 0
 
   try {
-    const chunks = splitChunksFast(fluxCSV)
+    fluxCSV = fluxCSV.trimEnd()
+
+    if (fluxCSV === '') {
+      return {
+        table: newTable(0),
+        fluxGroupKeyUnion: [],
+        resultColumnNames: [],
+      }
+    }
+
+    // Split the response into separate chunks whenever we encounter:
+    //
+    // 1. A newline
+    // 2. Followed by any amount of whitespace
+    // 3. Followed by a newline
+    // 4. Followed by a `#` character
+    //
+    // The last condition is [necessary][0] for handling CSV responses with
+    // values containing newlines.
+    //
+    // [0]: https://github.com/influxdata/influxdb/issues/15017
+
+    // finds the first non-whitespace character
+    let curr = fluxCSV.search(/\S/)
+
+    const chunks = []
+    while (curr !== -1) {
+      const oldVal = curr
+      const nextIndex = fluxCSV
+        .substring(curr, fluxCSV.length)
+        .search(/\n\s*\n#/)
+      if (nextIndex === -1) {
+        chunks.push([oldVal, fluxCSV.length])
+        curr = -1
+        break
+      } else {
+        chunks.push([oldVal, oldVal + nextIndex])
+        curr = oldVal + nextIndex + 2
+      }
+    }
 
     // declaring all nested variables here to reduce memory drain
     let tableText = ''
@@ -206,31 +245,57 @@ export const fastFromFlux = (fluxCSV: string): FromFluxResult => {
     let columnKey = ''
     let columnDefault: any = ''
 
-    for (const chunk of chunks) {
-      const splittedChunk = chunk.split('\n')
+    let chunk = ''
+    for (const [start, end] of chunks) {
+      let index = 0
+      chunk = fluxCSV.substring(start, end)
 
-      const tableTexts = []
-      const annotationTexts = []
+      annotationText = ''
+      tableText = ''
 
-      splittedChunk.forEach(line => {
-        if (line.startsWith('#')) {
-          annotationTexts.push(line)
+      while (index !== -1) {
+        const oldIndex = index
+        index = chunk.substring(oldIndex, chunk.length).search(/\n/)
+
+        if (index === -1) {
+          if (chunk[oldIndex] === '#') {
+            annotationText = `${annotationText}${chunk.substring(
+              oldIndex,
+              chunk.length
+            )}`
+          } else {
+            tableText = `${tableText}${chunk.substring(oldIndex, chunk.length)}`
+          }
+          break
         } else {
-          tableTexts.push(line)
+          if (chunk[oldIndex] === '#') {
+            annotationText = `${annotationText}${chunk.substring(
+              oldIndex,
+              oldIndex + index + 1
+            )}`
+          } else {
+            tableText = `${tableText}${chunk.substring(
+              oldIndex,
+              oldIndex + index + 1
+            )}`
+          }
+          index = index + oldIndex + 1
         }
-      })
-
-      tableText = tableTexts.join('\n').trim()
+      }
 
       assert(
         !!tableText,
         'could not find annotation lines in Flux response; are `annotations` enabled in the Flux query `dialect` option?'
       )
 
-      // TODO(ariel): csvParse is a slow operation
-      tableData = csvParse(tableText)
-
-      annotationText = annotationTexts.join('\n').trim()
+      /**
+       * csvParse is a slow operation, but so we may want to see whether we can
+       * find an alternative solution to the problem.
+       *
+       * Also, the trimStart here is because we're getting a \n prepended at some point,
+       * so this trims that off
+       */
+      tableData = csvParse(tableText.trimStart())
 
       assert(
         !!annotationText,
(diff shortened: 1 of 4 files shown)


giraffe/src/utils/fromFlux.ts
```diff
@@ -189,6 +189,162 @@ export const fromFlux = (fluxCSV: string): FromFluxResult => {
...
+
+          if (columnType === 'number') {
+            if (value === '') {
+              result = null
+            }
+            const parsedValue = Number(value)
+            result = parsedValue === parsedValue ? parsedValue : value
+          }
+
+          columns[columnKey].data[tableLength + i] = result
```
asalem1, line 310: we sacrifice readability for memory allocation improvements by removing the `parseValue` usage in the function and simply inlining the operation

giraffe/src/utils/fromFlux.ts
```diff
@@ -219,6 +375,51 @@ const splitChunks = (fluxCSV: string): string[] => {
...
+    if (nextIndex === -1) {
+      chunks.push([oldVal, trimmedResponse.length])
+      curr = -1
+      break
+    } else {
+      chunks.push([oldVal, oldVal + nextIndex])
+      curr = oldVal + nextIndex + 3
+    }
+  }
+  chunks = chunks.map(([start, end]) => trimmedResponse.substring(start, end))
```
asalem1, line 418: this operation uses the same `regex` pattern to identify the position of the chunks and saves a position reference to the chunk beginning and end, using those to then create the array of chunks based on the original CSV, rather than the more memory-intensive `split().map()` approach in `splitChunks`

subirjolly:
The functionality seems a lot faster but there's still that hang issue. It's not as severe though. But awesome improvement!

TCL735:
Should add a version number bump so that publishing won't need another commit.

TCL735 (approved):

giraffe/src/utils/fromFlux.ts
```diff
@@ -219,6 +364,49 @@ const splitChunks = (fluxCSV: string): string[] => {
   return chunks
 }
 
+/*
+  A Flux CSV response can contain multiple CSV files each joined by a newline.
+  This function splits up a CSV response into these individual CSV files.
+  See https://github.com/influxdata/flux/blob/master/docs/SPEC.md#multiple-tables.
+*/
+const splitChunksFast = (fluxCSV: string): string[] => {
+  const trimmedResponse = fluxCSV.trim()
```
drdelambre, line 373: duplicating data here

giraffe/src/utils/fromFlux.ts
```diff
@@ -219,6 +364,49 @@ const splitChunks = (fluxCSV: string): string[] => {
...
+    if (nextIndex === -1) {
+      chunks.push([oldVal, trimmedResponse.length])
+      curr = -1
+      break
+    } else {
+      chunks.push([oldVal, oldVal + nextIndex])
+      curr = oldVal + nextIndex + 2
+    }
+  }
+  return chunks.map(([start, end]) => trimmedResponse.substring(start, end))
```
drdelambre, line 407: duplicating data here

giraffe/src/utils/fromFlux.ts
```diff
@@ -189,6 +189,151 @@ export const fromFlux = (fluxCSV: string): FromFluxResult => {
...
 }
 
+export const fastFromFlux = (fluxCSV: string): FromFluxResult => {
+  const columns: Columns = {}
+  const fluxGroupKeyUnion = new Set<string>()
+  const resultColumnNames = new Set<string>()
+  let tableLength = 0
+
+  try {
+    const chunks = splitChunksFast(fluxCSV)
```
drdelambre, line 199: pull this out of the split function and back into the main function so you dont have the function overhead of passing these chunks around

giraffe/src/utils/fromFlux.ts
asalem1: Is there a way to have `trim` apply changes to the string without needing to be assigned onto a new string?

giraffe/src/utils/fromFlux.ts
asalem1: What alternative would you suggest here?

giraffe/src/utils/fromFlux.ts
asalem1: Gotcha

giraffe/src/utils/fromFlux.ts
```diff
@@ -189,6 +189,151 @@ export const fromFlux = (fluxCSV: string): FromFluxResult => {
...
+
+    // declaring all nested variables here to reduce memory drain
+    let tableText = ''
+    let tableData: any = []
+    let annotationText = ''
+    let columnType: any = ''
+    let columnKey = ''
+    let columnDefault: any = ''
+
+    for (const chunk of chunks) {
```
drdelambre, line 209: should be operating on indices here and not on the malloc'd chunks. eg: instead of splittedChunk = chunk.split('\n'), use the same iteration technique from the while loop that split up chunks, to split up newlines as well. iirc, because of json and the like, you only want to use this technique for finding the annotation header information before the results, and use papaparse for the rest of the response body

giraffe/src/utils/fromFlux.ts
```diff
@@ -189,6 +189,151 @@ export const fromFlux = (fluxCSV: string): FromFluxResult => {
...
+
+      splittedChunk.forEach(line => {
+        if (line.startsWith('#')) {
+          annotationTexts.push(line)
+        } else {
+          tableTexts.push(line)
+        }
+      })
+
+      tableText = tableTexts.join('\n').trim()
```
drdelambre, line 223: duplicated data by splitting 1000 lines as a string into 1000 objects containing a string, just to join the 1000 lines into a string again

giraffe/src/utils/fromFlux.ts
```diff
@@ -189,6 +189,151 @@ export const fromFlux = (fluxCSV: string): FromFluxResult => {
...
+
+      assert(
+        !!tableText,
+        'could not find annotation lines in Flux response; are `annotations` enabled in the Flux query `dialect` option?'
+      )
+
+      // TODO(ariel): csvParse is a slow operation
+      tableData = csvParse(tableText)
+
+      annotationText = annotationTexts.join('\n').trim()
```
drdelambre, line 233: same comment about splitting just to join again

asalem1:
> should be operating on indices here and not on the malloc'd chunks. eg: instead of splittedChunk = chunk.split('\n'), use the same iteration technique from the while loop that split up chunks, to split up newlines as well. iirc, because of json and the like, you only want to use this technique for finding the annotation header information before the results, and use papaparse for the rest of the response body

Totally hear you on these concerns and would love to address these in a future PR. Right now this PR aims to specifically target the most problematic area of the parser, make a small adjustment, derisk that move, then iterate on that. The comments here about what needs to be done are a great step forward w/r/t what needs to be done for the parser once we're able to prove that the existing chunks are a like for like replacement. Thoughts @drdelambre?

giraffe/src/utils/fromFlux.ts
drdelambre: instead of starting your index hunting while loop at `0`, start it at `n`, where `n` is determined by finding the first `\S` (non-whitespace character) instance

giraffe/src/utils/fromFlux.ts
```diff
@@ -196,7 +196,46 @@ export const fastFromFlux = (fluxCSV: string): FromFluxResult => {
...
+    // [0]: https://github.com/influxdata/influxdb/issues/15017
+
+    // finds the first non-whitespace character
+    let curr = fluxCSV.search(/\S/)
+
+    const chunks = []
+    while (curr !== -1) {
+      const oldVal = curr
+      const nextIndex = fluxCSV
+        .substring(curr, fluxCSV.length)
```
drdelambre, line 228: this is different than what you had inside the splitFastChunks function. why'd you change this logic? you're duping data here again and pretty egregiously (you're effectively duplicating the dataset every iteration). the only string you want to operate on is `fluxCSV` and use numbers, never operate on strings derived `fluxCSV` with numbers

giraffe/src/utils/fromFlux.ts
```diff
@@ -206,31 +245,57 @@ export const fastFromFlux = (fluxCSV: string): FromFluxResult => {
...
 
-      const tableTexts = []
-      const annotationTexts = []
+      annotationText = ''
+      tableText = ''
 
-      splittedChunk.forEach(line => {
-        if (line.startsWith('#')) {
-          annotationTexts.push(line)
+      while (index !== -1) {
```
drdelambre, line 256: note, this logic can be moved into the chunk discovery part after closing this PR

giraffe/src/utils/fromFlux.ts
```diff
@@ -206,31 +245,57 @@ export const fastFromFlux = (fluxCSV: string): FromFluxResult => {
...
+        const oldIndex = index
+        index = chunk.substring(oldIndex, chunk.length).search(/\n/)
+
+        if (index === -1) {
+          if (chunk[oldIndex] === '#') {
```
drdelambre, line 261: this is incorrect logic. you should move back to the old logic here and handle it in a different PR. the correct way iirc is to not split lines at all (and let papaparse do that to handle embedded json cases) and look for the first newline that is not followed by a `#`. everything before that point is annotation, every thing after is table text

PR #759, opened 2022-06-01
chore: clean up drift between parsers after proven success

asalem1:
This PR consolidates the changes that were made here:

https://github.com/influxdata/giraffe/pull/752

### Background

The linked PR was made to help address some performance issues with the fromFlux parser. After merging the changes in, publishing a new version of giraffe in the UI, and feature flagging the usages of fromFlux in the UI to use the `fastFromFlux` the parser changes appear to have improved the UI's usage of the parser without any known issues. Having said that, this implementation still relies upon allocating memory for chunks of data that are then parsed, chunked and joined again. While this imperfect implementation is a lot better than the previous one, we want to be able to iterate on the `fastFromFlux` parser without compromising gains made to the parser.

### Reviewer Notes

This PR basically just updates the changes that were applied to the `fastFromFlux` parser to the `fromFlux` parser after it's been a proven success. The plan here is to publish a new version of giraffe, turn the `fastFromFlux` feature flag off in the UI with the newly published version of giraffe, and iterate on updating the `fastFromFlux` parser so that we can improve performance without compromising gains

### DIFF
--- a/giraffe/src/utils/fromFlux.ts
+++ b/giraffe/src/utils/fromFlux.ts
@@ -77,7 +77,51 @@ export const fromFlux = (fluxCSV: string): FromFluxResult => {
   let tableLength = 0
 
   try {
-    const chunks = splitChunks(fluxCSV)
+    /*
+      A Flux CSV response can contain multiple CSV files each joined by a newline.
+      This function splits up a CSV response into these individual CSV files.
+      See https://github.com/influxdata/flux/blob/master/docs/SPEC.md#multiple-tables.
+    */
+    fluxCSV = fluxCSV.trimEnd()
+
+    if (fluxCSV === '') {
+      return {
+        table: newTable(0),
+        fluxGroupKeyUnion: [],
+        resultColumnNames: [],
+      }
+    }
+
+    // Split the response into separate chunks whenever we encounter:
+    //
+    // 1. A newline
+    // 2. Followed by any amount of whitespace
+    // 3. Followed by a newline
+    // 4. Followed by a `#` character
+    //
+    // The last condition is [necessary][0] for handling CSV responses with
+    // values containing newlines.
+    //
+    // [0]: https://github.com/influxdata/influxdb/issues/15017
+
+    // finds the first non-whitespace character
+    let curr = fluxCSV.search(/\S/)
+
+    const chunks = []
+    while (curr !== -1) {
+      const oldVal = curr
+      const nextIndex = fluxCSV
+        .substring(curr, fluxCSV.length)
+        .search(/\n\s*\n#/)
+      if (nextIndex === -1) {
+        chunks.push([oldVal, fluxCSV.length])
+        curr = -1
+        break
+      } else {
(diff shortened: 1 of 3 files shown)


giraffe/src/utils/fromFlux.ts
```diff
@@ -148,10 +194,35 @@ export const fromFlux = (fluxCSV: string): FromFluxResult => {
...
+          } else if (columnType === 'number') {
+            if (value === '') {
+              result = null
+            } else {
+              const parsedValue = Number(value)
+              result = parsedValue === parsedValue ? parsedValue : value
+            }
+          } else {
+            result = null
+          }
```
hoorayimhelping, line 223: It looks like this `else` block is not necessary, since `result` is defaulted to `null` on line 198.

giraffe/src/utils/fromFlux.ts
```diff
@@ -196,6 +267,11 @@ export const fastFromFlux = (fluxCSV: string): FromFluxResult => {
   let tableLength = 0
 
   try {
+    /*
+      A Flux CSV response can contain multiple CSV files each joined by a newline.
+      This function splits up a CSV response into these individual CSV files.
+      See https://github.com/influxdata/flux/blob/master/docs/SPEC.md#multiple-tables.
+    */
```
hoorayimhelping, line 274: Helpful, thank you :+1:

hoorayimhelping (approved):
Have a minor suggestion on variable naming for consistency

hoorayimhelping (changes requested):
Actually, before you merge this in, would you please update the version in `giraffe/giraffe/package.json` and `giraffe/stories/package.json`?

TCL735:
Generally, I like that the changes make the function more performant.

I am trying to fully understand all of the proposed changes that will go back into the original `fromFlux` function rather than just being its own separate function, so my review will take a bit of time to be completed. I may add comments or questions over a period of time rather than all at once.

Once I am completely done with comments and questions, I will clearly indicate so.

Thanks for your patience.

giraffe/src/utils/fromFlux.ts
```diff
@@ -77,7 +77,51 @@ export const fromFlux = (fluxCSV: string): FromFluxResult => {
   let tableLength = 0
 
   try {
-    const chunks = splitChunks(fluxCSV)
+    /*
+      A Flux CSV response can contain multiple CSV files each joined by a newline.
+      This function splits up a CSV response into these individual CSV files.
+      See https://github.com/influxdata/flux/blob/master/docs/SPEC.md#multiple-tables.
+    */
+    fluxCSV = fluxCSV.trimEnd()
```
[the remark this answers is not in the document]
asalem1: That's a great question. Ideally we wouldn't even want to use trimEnd. All of these operations are allocation memory to data that we already have. While this had previously been `trim`, we transitioned to using `trimEnd` while also using an index reference to find the first non-whitespace character and just start iterating from there (line 108)

TCL735 (approved):

PR #760, opened 2022-06-02
chore: reduce mallocs in favor and speed parser up

asalem1:
This implementation is currently a WIP to try and reduce the memory allocation operations that the parser is currently using to operate. This **does not** fully operate off of pointers, nor does it address the original chunking behavior (of searching for chunks based on searching for a regex match. 

Ideally we can use this as a blueprint for performing annotation and table text building as we define the boundaries of the chunk in the previous step and consolidate some of the logic so that we can eventually stream data.

### DIFF
--- a/giraffe/src/utils/fromFlux.ts
+++ b/giraffe/src/utils/fromFlux.ts
@@ -333,7 +333,8 @@ export const fastFromFlux = (fluxCSV: string): FromFluxResult => {
           fluxCSV[index - 1] === '\n' || fluxCSV[index - 1] === undefined
         // check to see if the one before that is whitespace or a new line
         const isWhitespaceOrNewLine =
-          fluxCSV[index - 2] === undefined || fluxCSV[index - 2]?.trim() === ''
+          fluxCSV[index - 2] === undefined ||
+          fluxCSV[index - 2]?.search(/\S/) === -1
         if (startsWithHash && isPrevNewLine && isWhitespaceOrNewLine) {
           const nextIndex =
             fluxCSV.substring(index, end).lastIndexOf('\n#') + index
--- a/giraffe/src/utils/fromFlux.ts
+++ b/giraffe/src/utils/fromFlux.ts
@@ -322,45 +322,68 @@ export const fastFromFlux = (fluxCSV: string): FromFluxResult => {
     let columnDefault: any = ''
 
     for (const [start, end] of chunks) {
-      const splittedChunk = fluxCSV.substring(start, end).split('\n')
+      annotationText = ''
+      tableText = ''
 
-      const tableTexts = []
-      const annotationTexts = []
+      let index = start
 
-      splittedChunk.forEach(line => {
-        if (line.startsWith('#')) {
-          annotationTexts.push(line)
+      while (index < end) {
+        const startsWithHash = fluxCSV[index] === '#'
+        const isPrevNewLine =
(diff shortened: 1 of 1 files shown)


giraffe/src/utils/fromFlux.ts
```diff
@@ -246,45 +246,68 @@ export const fastFromFlux = (fluxCSV: string): FromFluxResult => {
...
+        const startsWithHash = fluxCSV[index] === '#'
+        const isPrevNewLine =
+          fluxCSV[index - 1] === '\n' || fluxCSV[index - 1] === undefined
+        // check to see if the one before that is whitespace or a new line
+        const isWhitespaceOrNewLine =
+          fluxCSV[index - 2] === undefined || fluxCSV[index - 2]?.trim() === ''
+        if (startsWithHash && isPrevNewLine && isWhitespaceOrNewLine) {
+          const nextIndex =
+            fluxCSV.substring(index, end).lastIndexOf('\n#') + index
+          const endOfAnnotation = fluxCSV.indexOf('\n', nextIndex + 1)
```
asalem1, line 264: this still doesn't consider whether we're in the middle of a string or if this is actually the end of the annotation text

giraffe/src/utils/fromFlux.ts
```diff
@@ -246,45 +246,68 @@ export const fastFromFlux = (fluxCSV: string): FromFluxResult => {
...
+      while (index < end) {
+        const startsWithHash = fluxCSV[index] === '#'
+        const isPrevNewLine =
+          fluxCSV[index - 1] === '\n' || fluxCSV[index - 1] === undefined
+        // check to see if the one before that is whitespace or a new line
+        const isWhitespaceOrNewLine =
+          fluxCSV[index - 2] === undefined || fluxCSV[index - 2]?.trim() === ''
+        if (startsWithHash && isPrevNewLine && isWhitespaceOrNewLine) {
+          const nextIndex =
+            fluxCSV.substring(index, end).lastIndexOf('\n#') + index
```
asalem1, line 263: malloc ... i know

giraffe/src/utils/fromFlux.ts
```diff
@@ -246,45 +246,68 @@ export const fastFromFlux = (fluxCSV: string): FromFluxResult => {
...
+          const endOfAnnotation = fluxCSV.indexOf('\n', nextIndex + 1)
+
+          if (nextIndex === -1 || endOfAnnotation === -1) {
+            throw new Error(
+              'annotation text was incompelte and could not be processed'
+            )
+          }
+          // TODO(ariel): check to see if this is in the middle of a string
+          // or if this is actually the end of the annotation text
+          annotationText = fluxCSV.substring(index, endOfAnnotation)
```
asalem1, line 273: not sure if we want to store these indices as pointers and simply pass the annotation substring to the parser below

giraffe/src/utils/fromFlux.ts
```diff
@@ -246,45 +246,68 @@ export const fastFromFlux = (fluxCSV: string): FromFluxResult => {
...
+          index = endOfAnnotation + 1
         } else {
-          tableTexts.push(line)
+          tableText = fluxCSV.substring(index, end)
```
asalem1, line 276: same point as above

asalem1:
Closing this since the direction we should go is using Papaparse

PR #764, opened 2022-06-06
fix: allowed single stat to be selectable

david-rusnak:
- Changed `userSelect` CSS property to allow selectable single stats.

https://user-images.githubusercontent.com/106361125/172224249-a93a9472-c0a6-4a89-8089-cc1c625e7c77.mov


ChitlangeSahas:
If you can add a screen recording of it working, that will be great!

giraffe/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/giraffe",
-  "version": "2.26.1",
+  "version": "2.26.2",
```
ChitlangeSahas, line 3: 👍

ChitlangeSahas (approved):
Looks great! 

One suggestion: can we take this one step further and make it so that when we hover, the cursor changes to `I` that way it's more obvious that the text is selectable?

PR #769, opened 2022-06-27
chore: reconcile the fromFlux parser with the papaparse implementation

asalem1:
This PR is meant to clean-up the existing `fromFlux` function by applying the existing `fastFromFlux` logic to the `fromFlux` parser. The basis of this change integrates papaparse as the basis for the parser rather than making use of niche logic and d3-dsv. As a result, d3-dsv is now no longer necessary in giraffe and has been removed as a dependency. 

While the performance gains here are negligible, this implementation makes csv parsing a more straightforward implementation AND handles JSON data as expected, whereas the previous implementation failed to parse JSON data correctly and required several iterations of manual checking and built-in exceptions in order to address the issue

### Additional Information

The `fastFromFlux` changes have been live in all environments for 2 weeks with reports that previously broken implementations (JSON data in a csv) are now currently working, with no reported issues w/r/t data parsing throughout the UI

### DIFF
--- a/giraffe/src/utils/fromFlux.ts
+++ b/giraffe/src/utils/fromFlux.ts
@@ -103,17 +103,17 @@ export const fromFlux = (fluxCSV: string): FromFluxResult => {
 
     const chunks = []
     while (currentIndex !== -1) {
-      const oldVal = currentIndex
+      const prevIndex = currentIndex
       const nextIndex = fluxCSV
         .substring(currentIndex, fluxCSV.length)
         .search(/\n\s*\n#/)
       if (nextIndex === -1) {
-        chunks.push([oldVal, fluxCSV.length - 1])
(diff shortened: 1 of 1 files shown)


giraffe/src/utils/fromFlux.ts
```diff
@@ -111,7 +106,7 @@ export const fromFlux = (fluxCSV: string): FromFluxResult => {
       const prevIndex = currentIndex
       const nextIndex = fluxCSV
         .substring(currentIndex, fluxCSV.length)
-        .search(/\n\s*\n#(?=datatype|group|default)/)
+        .search(/\n\s*\n#/)
```
ChitlangeSahas, line 109: Why did we take away the logic that looks for the extra keywords?

drdelambre (approved):

ChitlangeSahas (approved):

PR #770, opened 2022-06-30
fix: fixes an issue where papaparser was handling csvs with returns weirdly

asalem1:
Related to https://github.com/influxdata/ui/issues/4928

### Background
Sometimes, annotated CSVs can end in a `\r\n` rather than the expected `\n`. While this format is valid, the parser gets a prepended `\n` that doesn't get trimmed, and which throws the entire parser off by pointing to the wrong started. In this case, this resulted in the headers not being set correctly, and therefore precluding data from the headers that should have been included

### Solution
MOAR POINTER MATH. Seriously though, the solution here is to update the chunk's starting point to the first non-whitespace character.

### Reviewer Notes
This PR includes tests to ensure the stability of the feature. It also rolls back the previous changes that were applied to the `fromFlux` parser since they introduced this breaking change. Finally, it removes `headers` as a concept that's stored on the `parsed` chunk since that's only relevant to us within the context of that if statement, so no need to store it elsewhere

### DIFF
--- a/giraffe/src/utils/fromFlux.ts
+++ b/giraffe/src/utils/fromFlux.ts
@@ -279,9 +279,9 @@ export const fastFromFlux = (fluxCSV: string): FromFluxResult => {
       See https://github.com/influxdata/flux/blob/master/docs/SPEC.md#multiple-tables.
     */
     // finds the first non-whitespace character
-    let curr = fluxCSV.search(/\S/)
+    let currentIndex = fluxCSV.search(/\S/)
 
-    if (curr === -1) {
+    if (currentIndex === -1) {
       return {
         table: newTable(0),
         fluxGroupKeyUnion: [],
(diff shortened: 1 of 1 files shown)


giraffe/src/utils/fromFlux.ts
```diff
@@ -307,7 +307,7 @@ export const fastFromFlux = (fluxCSV: string): FromFluxResult => {
       const oldVal = curr
       const nextIndex = fluxCSV
         .substring(curr, fluxCSV.length)
-        .search(/\n\s*\n#/)
+        .search(/\n\s*\n#(?=datatype|group|default)/)
```
ChitlangeSahas, line 310: Thank you for adding this 👍

drdelambre (approved):

ChitlangeSahas (approved):

PR #782, opened 2022-07-23
fix: update interface for SimpleTable and re-port implementation from influxdata ui

### LINKED ISSUE #777: Simple Table: update the interface to avoid re-parsing
Simple Table's interface should just take the `table` from the `result` similar to other visualizations such as the line graph. This allows Simple Table to avoid re-parsing the data in cases where the application already has the `result`. Let's make `table` the primary data source for Simple Table. It will be an optional property to avoid breaking third party apps who use Giraffe without the `ui`. But `ui` will consume Simple Table through this new property.

The reason is because the two implementations of Simple Table have drifted apart before we could finalize the design of Simple Table. The original work to port Simple Table to Giraffe was too rigid in implementing the interface. Let's correct this by allowing the interface to be more flexible.

Screen Shot 2022-07-18 at 4 38 55 PM

### LINKED ISSUE #781: Bug: SimpleTable: updating the results does not re-render the graph correctly
Updating the results given to SimpleTable's `PaginationProvider` its children does not cause the SimpleTable to re-render properly with the new data.

The video shows an example of the bug:

A SimpleTable initially given blank csv, which is converted into results with an empty `table`, empty `fluxGroupKeyUnion`, and empty `resultColumnNames`. It is then given valid csv from a flux query. Notice that the SimpleTable does not re-render properly with the new data.

This is due to the `PaginationProvider` not setting the correct page with the new results. Notice that once the user navigates to a different page, the results render the new data.

https://user-images.githubusercontent.com/10736577/180322999-771c8c19-7da4-46ee-80ec-18b9f2179b0e.mp4

TCL735:
Closes #777  
Closes #781  

- Updates the interface for `SimpleTable` to avoid re-parsing unnecessarily
- Re-ports the implementation from https://github.com/influxdata/ui

Screen Shot 2022-07-22 at 6 07 54 PM


giraffe/src/types/index.ts
```diff
@@ -460,7 +462,6 @@ export interface ScatterLayerConfig {
 export interface SimpleTableLayerConfig {
   type: 'simple table' // do not refactor or restrict to LayerTypes.SimpleTable
   showAll: boolean
-  fluxResponse?: string
```
TCL735, line 463: Redundant. This is already in the outer (parent) `config` object.

giraffe/src/utils/newTable.ts
```diff
@@ -2,7 +2,7 @@ import {Table, ColumnType, ColumnData, Config, FluxDataType} from '../types'
 import {fromFlux} from './fromFlux'
 
 // Don't export me!
-class SimpleTable implements Table {
+class EmpiricalTable implements Table {
```
TCL735, line 5: This was very confusing. An internal `SimpleTable` class that implements `Table` which is widely used by many other graph types. In logs and during debugging, this could lead to a misunderstanding of what `SimpleTable` actually is.

Renamed for clarity.

giraffe/README.md
```diff
@@ -342,14 +342,24 @@ When using the comma separated values (CSV) from the Flux query as the `fluxResp
...
+
+- **fromFluxResult**: _Object. Optional and used only by the SimpleTable layer type._ An object containing the properties of the result of parsing **fluxResponse**. Using this property prevents **SimpleTable** from unnecessarily re-parsing the **fluxResponse**. When both are missing, `config` will create an empty data set without any parsing. Ignored by all other layer types, see below.
+
+- **table**: _Object. Optional._ The data structure for a layer obtained from parsing a **fluxResponse**. **Not used by the SimpleTable and TableGraph** layer types (SimpleTable and TableGraph, see above). When present, overrides **fluxResponse**. When missing, `config` will look for **fluxResponse** and parse it to create the **table**. When both are missing, `config` will create an empty **table** without any parsing. This property is a data object produced by [Giraffe's utility functions](#utility-functions). Houses the data and getters and setters for that data, for the `<Plot>`.
```
hoorayimhelping, line 357: great job doing the work to understand this enough to summarize it here. The different types of flux data aren't at all obvious when looking at this from a high level.

After reading the summary in this form, the obvious question is: all this complexity in different types and optional responses and different preferences depending on which response is being used necessary, or is it just incidental? Can we eventually move to a state where we have a single response type and consistent interface?

giraffe/src/components/SimpleTable/PagedTable.tsx
```diff
@@ -109,26 +109,28 @@ const subsetResult = (
...
 
   const tables: SubsetTable[] = []
-  let lastTable
+  let lastTable = ''
 
   // group by table id (series)
   for (let ni = 0; ni < size; ni++) {
     if (
-      `y${subset['result'][0].data[ni]}:t${subset['table'][0].data[ni]}` ===
+      `y${subset['result']?.[0]?.data?.[ni]}:t${subset['table']?.[0]?.data?.[ni]}` ===
```
hoorayimhelping, line 117: It's not necessary in this PR, and I know you didn't initially write this code, but let's make sure we clean up some of these frankly atrocious variable names like `ni`, `y`, and `t`. Part of what makes this code so difficult to understand is that the variables are so poorly named, and have so much syntax (like `[`, `?`, `.`, `]`, etc).

It might also make sense to pull out common expressions into a named variable at the start of the loop, e.g.

```ts
for (let ni = 0; ni < size; ni++) {
const currentResult = subset['result']?.[0]?.data?.[ni]
const currentTable = subset['table']?.[0]?.data?.[ni]
```
(or some obviously better variable name if they're not actually the current result or current table. To my point about the variable names making this code really hard to follow, it's hard to even know what these things represent without diving deep into the code).

giraffe/src/utils/normalizeConfig.ts
```diff
@@ -16,7 +17,45 @@ export const normalizeLayers = (layers: LayerConfig[]): LayerConfig[] =>
...
-  layers: normalizeLayers(config?.layers),
-})
+export const normalizeConfig = (config: Config): Config => {
+  const dataProperties: Partial<Config> = {}
+
+  if (config) {
+    const {fluxResponse, fromFluxResult, table} = config
+
+    switch (config.layers?.[0]?.type) {
+      case 'simple table':
```
hoorayimhelping, line 27: Similar to how we put brackets around all `if` statements, even if it's not strictly necessary to guard against bugs, we should start getting into the habit of putting brackets around all `case` statements.

```ts
case 'simple table': {
  if (fromFluxResult) {
    ...
  }
  break
}

hoorayimhelping (approved):
LGTM! Great work digging into the knot and untangling a big part of it!! I really appreciate that you have the patience and hardheaded will to take on this kind of work and see it through.

One comment about brackets in `switch` statements and a couple of questions about future clean up work

giraffe/README.md
TCL735: Incidental.

Giraffe was originally part of the UI application. As we have slowly moved the visualization responsibilities from UI to Giraffe, the interface between them has evolved. On top of that, new visualization types were added during this transition, complicating the interface further.

The different ways of passing data to Giraffe should be viewed as flexibility rather than inconsistency. We were supporting open source users of Giraffe who write their consuming UI application differently than our own UI, and of course our own UI. The different interfaces exist to keep backwards compatibility.

PR #802, opened 2022-09-10
fix: speeds up height calculations for Simple Table

TCL735:
Closes https://github.com/influxdata/ui/issues/5692

By changing the initial row height from 0 to a positive number, we dramatically speed up the height estimation used by Simple Table during rendering of large data sets. The actual row height will auto-correct after the initial guess.

https://user-images.githubusercontent.com/10736577/189460829-093e593e-0ca1-4b7f-94bd-c05a0233c419.mp4

### DIFF
--- a/giraffe/package.json
+++ b/giraffe/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/giraffe",
-  "version": "2.34.1",
+  "version": "2.34.2",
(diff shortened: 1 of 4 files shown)


ChitlangeSahas (approved):
Thank you so much for taking care of this!

hoorayimhelping (approved):

Issue #805, opened 2022-09-14
SimpleTable: performance enhancement


TCL735:
On a large data set, Simple Table still performs poorly. While rendering eventually happens, there is a long pause (longer than for a line graph as a comparison) and the data finally renders. But the actual rendering could be very small, just a handful of rows.

It should not take _that_ long to render a subset of the data, as SimpleTable has built-in pagination. Something is clearly amiss with its internal implementation that causes a performance bottleneck.

Possible explanations:
- Sorting is happening on the entire data set, which can be n^2 or worse
- Unnecessary traversals through the entire data set are happening

TCL735:
Sorting on the entire data set does not appear to be happening. I scrambled a large data set and the rendering came back in the same order as the scrambled data.

However, there may still be many smaller sorts happening.

TCL735:
For the most part, Simple Table performance is about as good as it gets. We have addressed the `usePlotEnv` hook (it has been removed from Simple Table's code path), and also various major bugs.

Since this issue is about performance, I am closing it as done.

Additional work on Simple Table can be addressed in separate tickets.

PR #808, opened 2022-09-19
feat: memoize SimpleTableLayer and clean up its hooks

### LINKED ISSUE #806: SimpleTable: new data is not paginated correctly
When Simple Table renders as a blank table, due to no data, and then when new data is given, the updated rendering is not paginated correctly.

This can be reproduced in Storybook by using the Simple Table Custom CSV option. See the video below. Notice that the table is initially blank, because there is no data in the input field. Pasting valid CSV into the input field updates the rendering. However, the rendering is not paginated correctly. Every row is put on a single page, and any row that does not fit the height is hidden below the visible part of graph.

https://user-images.githubusercontent.com/10736577/190527270-d3b93bb9-b4d5-4e24-a106-c61a3443cc71.mp4

TCL735:
Closes #806  
Ongoing effort for #805  

Here is a video showing the data set being updated and then paginating correctly. Also, the large data set renders well.

https://user-images.githubusercontent.com/10736577/191131202-a8e38677-e94e-4e8c-ba6c-2a92f96c583d.mp4

### DIFF
--- a/giraffe/src/components/SimpleTable/PagedTable.tsx
+++ b/giraffe/src/components/SimpleTable/PagedTable.tsx
@@ -267,8 +267,7 @@ const PagedTable: FC<Props> = ({result, showAll}) => {
       pagedTableHeaderRef?.current?.children?.[0]?.clientHeight > 0 &&
       tableHeaderHeight === INITIAL_HEADER_HEIGHT
     ) {
-      const calculatedHeaderHeight =
-        pagedTableHeaderRef.current.clientHeight ?? 0
+      const calculatedHeaderHeight = pagedTableHeaderRef.current.clientHeight
 
       if (calculatedHeaderHeight !== tableHeaderHeight) {
         setTableHeaderHeight(calculatedHeaderHeight)
@@ -283,7 +282,7 @@ const PagedTable: FC<Props> = ({result, showAll}) => {
       tableRowHeight === INITIAL_ROW_HEIGHT
     ) {
       const calculatedRowHeight =
-        pagedTableBodyRef.current.children?.[0]?.clientHeight ?? 0
+        pagedTableBodyRef.current.children[0].clientHeight
 
       if (calculatedRowHeight !== tableRowHeight) {
         setTableRowHeight(calculatedRowHeight)
--- a/giraffe/src/components/SimpleTable/PagedTable.tsx
+++ b/giraffe/src/components/SimpleTable/PagedTable.tsx
@@ -268,7 +268,7 @@ const PagedTable: FC<Props> = ({result, showAll}) => {
   // eslint-disable-next-line react-hooks/exhaustive-deps
   useEffect(() => {
     if (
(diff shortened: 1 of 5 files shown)


giraffe/src/components/SimpleTable/PagedTable.tsx
```diff
@@ -230,18 +241,19 @@ const subsetResult = (
...
-  properties: SimpleTableViewProperties
+  showAll: boolean
   result: FluxResult['parsed']
 }
 
 const INITIAL_HEADER_HEIGHT = 0
 const INITIAL_HEIGHT = 0
-const INITIAL_ROW_HEIGHT = 10 // must be greater than 0
-const PagedTable: FC<Props> = ({result, properties}) => {
+const INITIAL_ROW_HEIGHT = 0
```
TCL735, line 250: The initial row height is indeed 0. The calculations that estimate the number of rows should handle its own logic in not using a 0 height to cause an infinite number of possible rows (and thus running through the entire data set). Thus, what we're doing here is moving the "must be greater than 0" logic to the correct place rather than arbitrarily putting it into this constant.

giraffe/src/components/SimpleTable/PagedTable.tsx
```diff
@@ -112,13 +123,13 @@ const subsetResult = (
       })
     )
     .filter(column => !!column.data.filter(_c => _c !== undefined).length)
-    .reduce((arr, curr) => {
-      if (arr[curr.name]) {
-        arr[curr.name].push(curr)
-        return arr
+    .reduce((acc, curr) => {
```
hoorayimhelping, line 126: It's astounding, bordering on impressive how bad this code is. Between the ridiculous variable names, the unnecessary `map().filter().reduce()` pattern that could be combined into a single loop that is much conceptually simpler, and the sheer number of loops in this function, which is on a hot rendering path, it's just a complete mess. I bet that simplifying this function will increase performance significantly, cause it is pretty bad.

Let me know if you want help with it. No one should have to deal with this code alone.

hoorayimhelping (approved):

giraffe/src/components/SimpleTable/PagedTable.tsx
TCL735: The whole sequence of array functions start from `columns` inside of the `table` property which is the object representing all of the data. The `columns` are not that many. They `can` be many if cardinality is high, if I am understanding cardinality correctly. But on the other hand, with a lot of columns, comes a lot of scrolling horizontally for the user to see all of the columns. Likely, a user would not want to visualize something that requires him/her to scroll incessantly to the side just to see some data.

So, the amount of performance to be gained here would depend on the type of data the user wants to see. I would in most cases, the number of columns here don't make enough of a difference on the rendering performance.

We can certainly and should change this code to make it more readable. Using one pass through the columns is a good approach. But also, there's a lot of work being done on each pass. We will need to be clear on the steps to make it easy to follow.

Let's leave this part for a future change. This is not the last update to Simple Table.

PR #809, opened 2022-09-20
fix: change font loader from file-loader to url-loader

hoorayimhelping:
Helps with: https://github.com/influxdata/ui/issues/1673
Helps with:  https://github.com/influxdata/ui/issues/2629

In cloud environments, there are font loading errors that are thrown in the console on every page. These don't show up in development environments, and so this has been a historically difficult issue to reproduce.

I was finally able to do it by creating a minimal application that used Clockface and Giraffe. I was able to fix it by changing the webpack configuration of Giraffe to use `url-loader` instead of `font-loader` when loading webfonts. [This issue in webpack's repo](https://github.com/webpack/webpack/issues/7353) was extremely helpful.

### Before
before

### After
after

Testing in the UI. If you clicked on one of the loaded fonts in the browser network tab, there would be an error when loading it. After, no such error.

### Before
https://user-images.githubusercontent.com/146112/191370921-96601592-e288-49b3-8ef3-bb4683bb0875.mov

### After
https://user-images.githubusercontent.com/146112/191370909-7d8b95ae-3edb-4a4f-8b51-85501e52517b.mov

### DIFF
--- a/giraffe/package.json
+++ b/giraffe/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/giraffe",
-  "version": "2.34.4",
+  "version": "2.34.5",
   "main": "dist/index.js",
   "module": "dist/index.js",
   "license": "MIT",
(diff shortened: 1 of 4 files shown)


ChitlangeSahas (approved):

ChitlangeSahas:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

TCL735 (approved):

TCL735:
Sweeeeet. The font loading issue can finally be put to rest.

PR #812, opened 2022-09-21
fix: change fonts on gauge and single-stat to Proxima Nova

### LINKED ISSUE #811: Giraffe: Gauge and SingleStat use a serif font
Here is the current Single Stat font:

Image

This is incorrect, as the font is specified as Rubik, which is also incorrect. It should be Proxima Nova

hoorayimhelping:
Previously, these fonts used Rubik, which doesn't exist in this repo anymore, which caused the fonts to fallback to the browser default, which appeared to be Times New Roman, or a generic serif font.

Closes #811

### After:
Screen Shot 2022-09-21 at 10 32 18 AM

Screen Shot 2022-09-21 at 10 32 15 AM


giraffe/src/constants/singleStatStyles.ts
```diff
@@ -5,25 +5,26 @@ export const LASER = '#00C9FF'
...
-  height: '100%',
-  padding: '8px',
+  borderRadius: '4px',
+  bottom: 0,
   color: LASER,
-  userSelect: 'text',
-  WebkitUserSelect: 'text',
+  cursor: 'text',
+  display: 'flex',
+  fontFamily: '"Proxima Nova", Helvetica, Arial, Tahoma, Verdana, sans-serif',
```
hoorayimhelping, line 14: this is just alphabetized. the only actual change was this line.

blegesse-git (approved):

PR #813, opened 2022-09-23
fix: limit row generation at the outset of the table

asalem1:
This PR adds a condition to initially render one row for a table that has no rows yet available. As is, the UI currently renders rows until the DOM becomes responsive to those changes and sets the data accordingly. In some cases, that means that we're rendering 600+ rows, then rerendering the table with the initial 13 rows. Since the initial row number is 0, we want to set the returned row number to 1 so that only 1 is rendered at the outset, and we can then calculate the row height properly based on that and return the correct number of rows to render.


TCL735:
Thanks for the fix!

We will incorporate this change, including the comment, in https://github.com/influxdata/giraffe/pull/814

PR #814, opened 2022-09-23
fix: split out visualizations that do not use PlotEnv

TCL735:
Part of #805  

Certain visualizations do not need the usePlotEnv hook, and the PlotEnv component which is very expensive. PlotEnv was also incorrectly being used twice (!) for `SizedTable`. Remove all usages completely.

Here is a video showing all of the non-PlotEnv visualizations working with the new split

https://user-images.githubusercontent.com/10736577/192071610-93362d29-a864-4bc4-976d-3554ae778c6b.mp4

### DIFF
--- a/giraffe/src/components/Plot.tsx
+++ b/giraffe/src/components/Plot.tsx
@@ -1,9 +1,16 @@
+// Libraries
 import React, {FunctionComponent, RefObject, useRef} from 'react'
 import AutoSizer from 'react-virtualized-auto-sizer'
 
+// Components
+import {PlotResizer, TableResizer} from './PlotResizer'
+
+// Types
 import {Config, SizedConfig} from '../types'
 
-import {PlotResizer} from './PlotResizer'
+// Utils
+import {hasPlotEnv} from '../utils/hasPlotEnv'
+
 interface PlotProps {
   config: Config
   axesCanvasRef?: RefObject<HTMLCanvasElement>
@@ -25,7 +32,7 @@ export const Plot: FunctionComponent<PlotProps> = props => {
   }
 
   if (config.width && config.height) {
-    return (
+    return hasPlotEnv(config) ? (
       <div className="giraffe-fixedsizer" style={{position: 'relative'}}>
         <PlotResizer
           axesCanvasRef={axesCanvasRef}
@@ -37,10 +44,18 @@ export const Plot: FunctionComponent<PlotProps> = props => {
           {children}
         </PlotResizer>
       </div>
+    ) : (
+      <TableResizer
+        config={config as SizedConfig}
+        height={config.height}
+        width={config.width}
+      >
+        {children}
+      </TableResizer>
     )
   }
 
-  return (
+  return hasPlotEnv(config) ? (
     <AutoSizer className="giraffe-autosizer">
       {({width, height}) => (
         <PlotResizer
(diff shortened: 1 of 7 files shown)


giraffe/src/components/SizedTable.tsx
```diff
@@ -113,22 +105,6 @@ export const SizedTable: FunctionComponent<Props> = ({
...
-                  innerHeight: env.innerHeight,
-                  xScale: env.xScale,
-                  yScale: env.yScale,
-                  xDomain: env.xDomain,
-                  yDomain: env.yDomain,
-                  columnFormatter: env.getFormatterForColumn,
-                  yColumnType: env.yColumnType,
-                }
-                return layerConfig.render(renderProps)
-
```
TCL735, line 131: `SizedTable` no longer can have a custom layer. This change makes sense because putting a canvas on top of non-canvas components is asking for bugs.

Custom layers will be limited to being in `SizedPlot` only.

giraffe/src/components/Plot.tsx
```diff
@@ -37,10 +44,18 @@ export const Plot: FunctionComponent<PlotProps> = props => {
...
+        height={config.height}
+        width={config.width}
+      >
+        {children}
+      </TableResizer>
     )
   }
 
-  return (
+  return hasPlotEnv(config) ? (
```
hoorayimhelping, line 58: what is the relationship between `plotEnv` and `layerCanvasRef`? it seems like not having `plotEnv` means the `layerCanvasRef` isn't necessary? is that try? why or why not?

giraffe/src/components/SimpleTable/PagedTable.tsx
```diff
@@ -40,24 +38,24 @@ const getNumberOfRowsOnCurrentPage = (
...
-  const minimumLength = result?.table?.length ?? 1
-  const estimatedRowHeight = Math.min(
-    Math.ceil(totalAvailableHeight / minimumLength),
-    MAXIMUM_ESTIMATED_ROW_HEIGHT
-  )
-  const visibleRowHeight = Math.max(rowHeight, estimatedRowHeight)
+  // this means that no rows have been mounted or measured, so we need to
+  // mount one row to measure the row height
+  if (rowHeight <= 0) {
+    return 1
```
hoorayimhelping, line 48: this seems like a smell. It seems like an indication that this function is doing too much, or isn't doing what it says its doing when we have to return 1 when there is no row height. Seems like we have more refactoring to do, because to me, this doesn't make a lot of logical sense

hoorayimhelping (approved):
Generally looks good to me, but I'm a little wary of the inferrence between not needing a plotEnv and not needing a `layerCanvasRef`

giraffe/src/components/Plot.tsx
TCL735: The `layerCanvasRef` is passed directly to the layers that use it for rendering such as `lineLayer` and `rectLayer`.

The `plotEnv` deals with the state of a plot, such as its tick marks, margins, and current config which in turn affects its spec which is what different layer types use for their own state.

TCL735: Since tables do not use a canvas a `layerCanvasRef` is not necessary. `plotEnv` is also not necessary for tables because they have limited state that they handle on their own.

giraffe/src/components/SimpleTable/PagedTable.tsx
TCL735: There are further refactor changes pending.

This pull request addresses the issue of the `usePlotEnv` hook dramatically slowing down Simple Table which was found by @asalem1

PR #819, opened 2022-10-03
chore: use resolutions to use d3-color 3.1.0

jdstrand:
* chore: use resolutions to use d3-color 3.1.0
    
    d3-scale and d3-interpolate both specify "d3-color@1 - 3" and are
    already at 3.0.1 in the upstream yarn.lock, so it should be safe to use
    3.1.0 in a resolution since according to semver, 3.1 should not break
    3.0.
* package.json: use ^1.7.3 for shell-quote rather than exact version

    This resolves a yarn warning introduced in 4b61d6bf.

Note: I've not personally tested this. I've you'd like testing beyond the automatic tests from the PR submission, please advise.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
   "resolutions": {
     "immer": "^9.0.6",
     "jsprim": "^1.4.2",
-    "shell-quote": "1.7.3"
+    "shell-quote": "1.7.3",
+    "d3-color": "^3.1.0"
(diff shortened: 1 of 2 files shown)


hoorayimhelping:
seems reasonable, based on this: https://github.com/d3/d3-color/releases/tag/v3.1.0

PR #820, opened 2022-10-03
chore: update d3 deps to use newer versions for d3-color security update

jdstrand:
*  chore: update d3 deps to use newer versions for d3-color security update
    Update giraffe/packages.json to use:
    * d3-color ^3.1.0
    * d3-format ^3.0.1
    * d3-interpolate ^3.0.1
    * d3-scale ^4.0.2
    
    This resolves d3-color in devDependencies to 3.1.0 to fix a security
    issue.
* chore: package.json: use ^1.7.3 for shell-quote rather than exact version
    
    This resolves a yarn warning introduced in 4b61d6b.

IMPORTANT: I've not tested this beyond verifying CI tests pass. Please advise if you'd like me to test manually.


TCL735:
I am re-opening this pull request and noting that we need to do something to get the build to pass.

Right now, there is an error when running the jest tests on many of the files:
```
SyntaxError: Unexpected token 'export'
```

The solution appears to be: https://stackoverflow.com/questions/49263429/jest-gives-an-error-syntaxerror-unexpected-token-export

PR #821, opened 2022-10-06
fix: count correctly the number of table headers that are rendered

TCL735:
Closes https://github.com/influxdata/ui/issues/6004

Fixes the problem with the measurement of the number of rows that can fit in the available space. Table headers, which are taller than a row, were not being counted correctly. Thus, the total running height was underestimated, and the available remaining height was overestimated, which led to squeezing in an extra row that should not have been rendered.

BEFORE - notice the last row on page 2 overlaps with the edge of the table and horizontal scrollbar:

https://user-images.githubusercontent.com/10736577/194189458-51df473b-a582-4cab-ad24-0344cbd4f8f6.mp4

AFTER - fixed, no overlapping:

https://user-images.githubusercontent.com/10736577/194189491-d4a8250b-b2d8-4c4d-8e04-c28f9a6d4cb0.mp4


giraffe/src/components/SimpleTable/PagedTable.tsx
```diff
@@ -54,31 +56,28 @@ const getNumberOfRowsOnCurrentPage = (
...
-            )}`
-        )
+        .map(column => {
+          if (column.data[rowIdx] !== undefined) {
+            return `${column.name}::${
+              column.fluxDataType
+            }::${result.fluxGroupKeyUnion.includes(column.name)}`
+          }
+          return 'undefined'
+        })
```
TCL735, line 69: This is the heart of the fix. The signature, which is a string value representing a table header, did not change when a new table with a different table header should be counted. We fix this by accurately creating the signature to detect if Simple Table will create a new table header or not.

giraffe/src/components/SimpleTable/PagedTable.tsx
```diff
@@ -54,31 +56,28 @@ const getNumberOfRowsOnCurrentPage = (
   let lastSignature
   let signature
 
-  // rowHeight is now guaranteed to be greater than zero
-  const lastVisibleRowMinimumHeight = 0.2 * rowHeight
-
```
TCL735, line 59: `lastVisibleRowMinimumHeight` was not accurately descriptively of what this value was actually doing. This value was an _additional_ margin of space on top of the space already allocated to the next row. We don't need this.

Previously, we put in this value because we didn't realize the `signature` (see below) was not accurately detecting new table headers, which are taller than a row.

giraffe/src/components/SimpleTable/PagedTable.tsx
```diff
@@ -54,31 +56,28 @@ const getNumberOfRowsOnCurrentPage = (
...
-              c.name
-            )}`
-        )
+        .map(column => {
+          if (column.data[rowIdx] !== undefined) {
+            return `${column.name}::${
+              column.fluxDataType
+            }::${result.fluxGroupKeyUnion.includes(column.name)}`
+          }
+          return 'undefined'
```
wdoconnell, line 68: What's this portion of the code - returning the string `undefined` - doing?

hoorayimhelping: @wdoconnell it's building a 'signature' (what actually seems like a [composite key](https://en.wikipedia.org/wiki/Composite_key)) based on the `name`, `fluxDataType` and `fluxGroupUnionKey`. It looks like the previous code incorrectly didn't account for when `column.data[rowIdx]` didn't exist

giraffe/src/components/SimpleTable/PagedTable.tsx
```diff
@@ -54,31 +56,28 @@ const getNumberOfRowsOnCurrentPage = (
...
-            `${c.name}::${c.fluxDataType}::${result.fluxGroupKeyUnion.includes(
-              c.name
-            )}`
-        )
+        .map(column => {
+          if (column.data[rowIdx] !== undefined) {
+            return `${column.name}::${
+              column.fluxDataType
+            }::${result.fluxGroupKeyUnion.includes(column.name)}`
+          }
```
hoorayimhelping, line 67: already easier to follow! 👍

giraffe/src/components/SimpleTable/PagedTable.tsx
```diff
@@ -54,31 +56,28 @@ const getNumberOfRowsOnCurrentPage = (
...
+          return 'undefined'
+        })
         .join('|')
 
       if (signature !== lastSignature) {
         runningHeight += headerHeight
 
         if (currentTable !== undefined) {
-          runningHeight += 10
+          runningHeight += TABLE_MARGIN
```
hoorayimhelping, line 76: :+1:

hoorayimhelping (approved):

giraffe/src/components/SimpleTable/PagedTable.tsx
TCL735: Excellent explanation. Thanks @hoorayimhelping

PR #822, opened 2022-10-07
fix: put Single Stat default styles into stylesheet to prevent font flashing

TCL735:
Part of https://github.com/influxdata/ui/issues/5993

- Removes inline styles for Single Stat's default styling
- Uses a stylesheet to load Single Stat's font to prevent font flashing

BEFORE: notice the font changes thickness

https://user-images.githubusercontent.com/10736577/194672131-1006aea7-5f7c-4c0c-9222-0d4dd55a9892.mp4

AFTER: no change in thickness of the font

https://user-images.githubusercontent.com/10736577/194672152-726f3954-1e81-4485-8c6c-ecaa1eaa1f2f.mp4


hoorayimhelping (approved):
Nice work!

Only question: should the font be the heavy / bold version? do our designers haven an opinion on this?

TCL735:
> Nice work!
> 
> Only question: should the font be the heavy / bold version? do our designers haven an opinion on this?

Good question. I copied these same font settings from Clockface. i will make sure to update them when they are updated in Clockface.

PR #897, opened 2026-08-31
chore: override d3-color to ^3.1.0 and fix issues introduced in previous depbot PRs

jdstrand:
This updates `d3-color` for transitive dependencies to also use 3.1.0, fixing an unimportant dependabot alert. Instead of dismissing the alert, move forward so we can continue to maintain this repo. This required a change to `giraffe/jest.config.js` due to a change in `d3-color` 3.1.0

In testing this, it was discovered the PR #886 broke CI since it needed node >=18, but we only had 16 (for some reason, `build` wasn't required to pass and was missed). In working through that, it was discovered that `node-sass` is unsupported and needs node 18 or earlier, which is EOL. As we move forward with dependabot updates, dependencies will require node 20 or newer more and more, so use the supported `sass` and update to `cimg/node:24.20` (current `lts`; we don't use `lts` yet since it will move to `26` at some point, potentially breaking builds).

`webpack` v4 uses MD4 which is removed from openssl 3. Newer versions of node (after 14) use openssl 3, so pass `NODE_OPTIONS=--openssl-legacy-provider` to storybook (which calls webpack) until we can upgrade.

Run `yarn run build-storybook` in CI.


eatondustin1 (dismissed):

jdstrand:
> for some reason, `build` wasn't required to pass and was missed
I updated the `giraffe` circleci setting to require `build` to pass. Hopefully that will help going forward.

jdstrand:
@eatondustin1 - hey, can you re-review? I had to make a number of additional changes after realizing the repo wasn't passing `build` for a while. Thanks!

eatondustin1 (approved):

